PR #8, opened 2020-02-07
Add feature spec for creating a collection

### LINKED ISSUE #7: Create a test for creating a collection
This happens in argo and uses the registration service in dor-services-app

mjgiarlo:
Fixes #7 

Includes:
* Rejigger Capybara to allow multiple specs to run serially using the initial set of credentials.
* Extract authentication code to a spec helper module
* Remove debugging statement printing download path
* Rename preassembly feature spec to clarify its intention, and to signal that going forward we expect each feature spec to test a single feature.

Note that I ran the tests and can attest that they passed for me. It might be useful to have one other @sul-dlss/infrastructure-team member pull down this branch and verify that the specs pass for them as well.


spec/spec_helper.rb
```diff
@@ -28,11 +27,12 @@
 #   end
 # end
 
-Capybara.current_driver = :my_firefox_driver
+Capybara.default_driver = :my_firefox_driver
```
mjgiarlo: This change was required to allow multiple specs to use the same browser session.

spec/support/authentication_helpers.rb
```diff
@@ -0,0 +1,37 @@
...
+      print "SUNet ID: "
+      username = $stdin.gets
+      username.strip
+    end
+
+    @@password ||= begin
+      print "Password: "
+      password = $stdin.noecho(&:gets)
+      # So the user knows we're off the password prompt
+      puts
```
mjgiarlo: I know it looks weird to have the puts here, but if it's the last line in the begin/end block, it interferes with the return value. And if it's outside the block, there's unnecessary newlines in the spec output.

peetucket:
Pulled the branch and ran tests and it works for me.

spec/features/create_collection_spec.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+require 'io/console'
+require 'random_word'
+
+RSpec.describe 'Create a collection in Argo', type: :feature do
```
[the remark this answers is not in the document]
mjgiarlo: Made the change. thanks!

PR #10, opened 2020-02-12
Feature for creating an object without files

jermnelson:
Feature runs through process of registering and accessioning an object with no files. 

Closes: #6


mjgiarlo:
@jermnelson 
> Open question: Should we go through at least `accessioningWF` step for this object?

I think we should, like the other tests, yeah. :+1:

PR #27, opened 2020-03-05
Add an integration test for sdr-api and sdr-client

jcoyne:
## Why was this change made?

So that we can test deposited file.

n/a


README.md
```diff
@@ -13,11 +13,13 @@ This script depends on having firefox downloaded
 1. Uncompress it `tar -zxvf geckodriver-v0.26.0-macos.tar.gz`
 1. Put geckodriver on the path `export PATH=$PATH:$(pwd)`
 
+## Get credentials in SDR-API
```
justinlittman: Can you describe where to do this? Like "on sdr-api stage server"?

jcoyne: Doesn't the link below have sufficient information?  Do we need to improve it here or on that other page?

justinlittman: Here.

spec/features/sdr_deposit_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  let(:source_id) { "testing:#{SecureRandom.uuid}" }
+  let(:apo) { 'druid:sc012gz0974' }
+  let(:collection) { 'druid:hv320hk4091' }
+  let(:catkey) { '10065784' }
+
+  before do
+    authenticate!(start_url: start_url, expected_text: 'Welcome to Argo!')
+  end
+
+  # This allows a login using credentials from the config gem.
```
mjgiarlo: Is `the config gem` copypasta? I didn't think we used that gem in this codebase.

spec/features/sdr_deposit_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  class LoginFromEnv
+    def self.run
+      { email: ENV.fetch('SDR_EMAIL'), password: ENV.fetch('SDR_PASSWORD') }
+    end
+  end
+
+  it do
```
mjgiarlo: It might be nice to have a little text description of the expectation here, but the test itself is also pretty easy to read.

.rubocop_todo.yml
```diff
@@ -31,14 +31,20 @@ Metrics/BlockLength:
...
     - 'spec/features/create_apo_spec.rb'
     - 'spec/features/create_collection_spec.rb'
     - 'spec/features/create_object_no_files_spec.rb'
     - 'spec/features/reaccession_via_preassembly_spec.rb'
+    - 'spec/features/sdr_deposit_spec.rb'
+
+# Offense count: 1
+RSpec/LeakyConstantDeclaration:
+  Exclude:
+    - 'spec/features/sdr_deposit_spec.rb'
```
ndushay: looking at https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/LeakyConstantDeclaration this *seems* like something we should care about (but looking at the spec I don't understand exactly what it is caring about)

README.md
```diff
@@ -13,11 +13,15 @@ This script depends on having firefox downloaded
...
+
+Create credentials on SDR-API stage following these instructions:
+https://github.com/sul-dlss/sdr-api#create-a-user
 
 ## Run
 
 You must be on VPN in order to run this script.
 
-`bundle exec rspec`
+`SDR_EMAIL=jcoyne85@stanford.edu SDR_PASSWORD='sekret!' bundle exec rspec`
```
ndushay: the password is sent over https, so it is encrypted, yes?

.rubocop_todo.yml
mjgiarlo: I guess it's about the login helper class. Can get rid of this exclusion if the class's name is destroyed via `const_remove` in an `after` hook, if memory serves.

README.md
mjgiarlo: :+1:

.rubocop_todo.yml
ndushay: well, that's obtuse.

jcoyne: Not a concern as this is only temporary until we can proxy credentials

jcoyne: Not a concern as this is only temporary until we can proxy credentials

PR #33, opened 2020-03-18
Technical Metadata Service Test

### LINKED ISSUE #24: new techmd-service needs integration tests
~~Once it becomes visible in Argo (https://github.com/sul-dlss/argo/issues/1824).~~

Blocked by an sdr-api integration test

confirm that it occurs as part of accessioning AND that it does the "right thing".  Maybe something like this??
- API call to tech-md-service for not-yet-accessioned new object is empty
- object accessioned
- API call to tech-md-service for recently accessioned object created tech-md for preserved files and NOT for unpreserved files

jermnelson:
## Why was this change made?
Fixes #24 

Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,6 @@ This script depends on having firefox downloaded
 
 You must be on VPN in order to run this script.
 
-`bundle exec rspec`
+`TECHMD_HMAC={hash-value} bundle exec rspec`
 
 You will be prompted to put in your credentials and then to approve a duo push.
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 gem 'byebug'
 gem 'capybara'
+gem 'jwt'
 gem 'rake'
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -17,6 +17,6 @@ This script depends on having firefox downloaded
 
 You must be on VPN in order to run this script.
 
-`bundle exec rspec`
+`TECHMD_HMAC={hash-value} bundle exec rspec`
```
jcoyne: This is unnecessary, we can just look for the techMetadata in argo:  for example the "Technical Metadata" section on this page: https://argo-stage.stanford.edu/view/druid:jw234dx5637?beta=true

jermnelson: Happy to make the change but was following the suggestions in the ticket

jcoyne: The TechMd is not an external API, so I don't think we want our integration tests to be calling it.

jermnelson:
Decided after consulting with @justinlittman to close this PR in favor of adding to an existing test for checking technical metadata.

PR #34, opened 2020-03-20
Add specs for single and bulk tag operations in Argo

mjgiarlo:
Connects to sul-dlss/dor-services-app#679

And move common requires into the spec_helper.


Gemfile
```diff
@@ -13,4 +14,4 @@ gem 'rspec'
 gem 'rubocop', '~> 0.73.0'
 gem 'rubocop-rspec'
 gem 'selenium-webdriver'
-gem 'sdr-client', '~> 0.16.0'
+gem 'sdr-client', '~> 0.16'
```
mjgiarlo: @jcoyne OK to relax this?

ndushay: erg - isn't this how we end up breaking a ton of apps at once if a pre 1.0 release has backward breaking changes?

spec/features/edit_bulk_tags_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    find('span#paste-druids-button').click
+    expect(page).to have_content 'Bulk actions will be performed on this list of druids'
+    find('textarea#pids').fill_in(with: bulk_druids.join("\n"))
+    click_button 'Tags'
+    expect(page).to have_content 'Change tags'
+    find('textarea#tags').fill_in(with: druids_with_tags)
+    find('span#set_tags').click
+    expect(page).to have_content 'Done!'
+    expect(page).to have_content "Using #{bulk_druids.count} user supplied druids and tags"
+    expect(page).not_to have_content(/error/i)
```
ndushay: this is great, but you should also check that the tags are now on the objects, no?  And same for the different flavors of updating tags ...

Gemfile
mjgiarlo: that's when we pin DSC to the minor of cocina-models, I think. I wasn't too worried about our integration tests using the latest and greatest.

spec/spec_helper.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
-require 'capybara/rspec'
-require 'selenium-webdriver'
 require 'byebug'
+require 'capybara/rspec'
+require 'io/console'
 require 'pry-byebug'
+require 'random_word'
```
ndushay: 1. is this used in more than one place
2. where it is used ... does it need to be a random word?  

ok to make a ticket for this

Gemfile
ndushay: oh duh.  reviewing too many PRs at once.

spec/spec_helper.rb
mjgiarlo: @ndushay yeah, it's used in 7 of our 9 feature specs. It's used to create random-looking text strings, such as titles and tags. It's more useful than hard-coding e.g. "Test Collection Title" because it gives us a string that we can look for later that isn't confused with another object.

ndushay: cool - thanks.

spec/features/edit_bulk_tags_spec.rb
mjgiarlo: @ndushay All of that is tested in `edit_item_tags_spec.rb`. This test stops at the edge of the bulk tags UI interaction. And bulk tags are not automatically reindexed, so it didn't seem worth the effort to take this test any farther at the time.

ndushay: ok.

PR #37, opened 2020-03-25
Etd test

### LINKED ISSUE #17: etd accessioning needs integration tests
from as early in the ETD processing as we get involved, if possible.   Would be great if we could start with hydra_etd and somehow mock the interactions with registrars office, etc.

If nothing else, we could start with when our own etd code passes the object on to accessioning?  This would be a NEW object and its characteristics could at least be created to mimic what we get from etd UI?

also, if it is done: update existing ETD object (whatever might happen)

See https://github.com/sul-dlss/hydra_etd/wiki/End-to-End-Testing-Procedure

ndushay:
## Why was this change made?

Closes #17

Added integration test for ETDs.  

#### What it does
- starts with a fake POST from the registrar that creates an ETD object in the hydra_etd app AND in fedora
- goes through getting all the necessary approvals and documents thru step 6 from https://github.com/sul-dlss/hydra_etd/wiki/End-to-End-Testing-Procedure up 
- checks Argo after registrar approval step to see the object has the right data and the etdSubmitWF is in the expected state

#### What it doesn't do  
- fake WF steps that have to do with Symphony and/or cron jobs (submit-marc, check-marc, catalog-status)
- run the ETD through common-accessioning (as this would happen after the aforementioned steps)
- check Argo after any of these non-done steps to prove the not-done steps are done.

na


spec/features/create_etd_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+      <career code="MED">Medicine</career>
+      <program code="MED">Medical</program>
+      <plan code="ANT">Neurology</plan>
+      <degree>PHD</degree>
+    </DISSERTATION>
+    XML
+  end
+
+  # may need to increase this: file uploads and submission to registrar
+  # Capybara.default_max_wait_time = 5
```
jcoyne: Can this be removed? These settings are in spec_helper

spec/features/create_etd_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+
+  # See https://github.com/sul-dlss/hydra_etd/wiki/End-to-End-Testing-Procedure
+  it do
+    # registrar creates ETD in hydra_etd application by posting xml
+    resp_body = simulate_registrar_post(xml_from_registrar)
+    prefixed_druid = resp_body.split.first
+    expect(prefixed_druid).to start_with('druid:')
+    puts "dissertation id is #{dissertation_id}"
```
jcoyne: Remove these puts please

spec/features/create_etd_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+    #   I think this may be an artifact of the js framework it usees, prototype
+    file_upload_elements = all('input[type=file]', visible: false)
+
+    # upload dissertation PDF
+    filename = 'etd_dissertation.pdf'
+    expect(page.find('#pbDissertationUploaded')['style']).to eq '' # dissertation PDF not yet provided
+    expect(page).not_to have_content(filename)
+    dissertation_pdf_upload_input = file_upload_elements.first
+    dissertation_pdf_upload_input.attach_file("spec/fixtures/#{filename}")
+    sleep(3) # wait for upload
```
jcoyne: This shouldn’t be required with Capybara. It waits on the next assertion until it’s true or there is a timeout. Though I think you need to flip the assertion to be a capybara find.

spec/features/create_etd_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+    expect(page).to have_content(filename)
+    # a checked box in the progress section is a background image
+    expect(page.find('#pbDissertationUploaded')['style']).to match(/background-image/)
+
+    # upload supplemental file
+    find('input#cbSupplementalFiles').check
+    filename = 'etd_supplemental.txt'
+    expect(page).not_to have_content(filename)
+    # supplemental files uploaded progress checkbox not visible by default
+    expect(page.find('#pbSupplementalFilesUploaded', visible: false)).not_to be_visible
```
jcoyne: This seems redundant

spec/features/create_etd_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+    # provide copyright permissions letters/files
+    # permission files uploaded progress checkbox not visible by default
+    expect(page.find('#pbPermissionFilesUploaded', visible: false)).not_to be_visible
+    filename = 'etd_permissions.pdf'
+    expect(page).not_to have_content(filename)
+    permissions_upload_input = file_upload_elements[11]
+    permissions_upload_input.attach_file("spec/fixtures/#{filename}")
+    expect(page).to have_content(filename)
+    # a checked box in the progress section is a background image
+    expect(page.find('#pbPermissionsProvided')['style']).to match(/background-image/)
```
jcoyne: Can you use have_css selectors?

spec/features/create_etd_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+    expect(page.find('#pbPermissionFilesUploaded')).to be_instance_of Capybara::Node::Element
+
+    # apply licenses
+    expect(page.find('#pbRightsSelected')['style']).to eq '' # rights not applied yet
+    click_link('View Stanford University publication license')
+    page.find('input#cbLicenseStanford').check
+    click_link('Close this window')
+    click_link('View Creative Commons licenses')
+    page.select 'CC Attribution license', from: 'selectCCLicenseOptions'
```
jcoyne: Page should be unnecessary

spec/features/create_etd_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+    click_link('Close this window')
+
+    # set embargo
+    click_link('Postpone release')
+    page.select '6 months', from: 'selectReleaseDelayOptions'
```
jcoyne: Page is unnecessary

spec/features/create_etd_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+    click_link('Close this window')
+
+    expect(page.find('#pbRightsSelected')['style']).to match(/background-image/)
+
+    # "submit etd to registrar"
+    accept_alert do
+      page.find('#submitToRegistrar').click # javascript
```
jcoyne: Can you use click_button?

spec/features/create_etd_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+    end
+    # page.find waits for this element to appear
+    expect(page.find('#submissionSuccessful')).to have_content('Submission successful')
```
jcoyne: Use have_selector with text:

spec/features/create_etd_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+    expect(page.find('#submitToRegistrarDiv > p.progressItemChecked')).to have_content('Submitted')
+
+    # fake reader approval
+    reader_progress_list_el = all('#progressBoxContent > ol > li')[8]
+    expect(reader_progress_list_el).to have_content('Verified by Final Reader - Not done')
+    reader_approved = xml_from_registrar.dup.sub(/<readerapproval>Not Submitted<\/readerapproval>/,
```
jcoyne: Can this be extracted to a let or a method?

spec/features/create_etd_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+                                                    '<regapproval>Approved</regapproval>')
+    now = Time.now.strftime('%m/%d/%Y %T')
+    registrar_approved.sub!(/<regactiondttm> <\/regactiondttm>/, "<regactiondttm>#{now}</regactiondttm>")
+    resp_body = simulate_registrar_post(registrar_approved)
+    expect(resp_body).to eq "#{prefixed_druid} updated"
+    page.refresh # needed to show updated progress box
+    registrar_progress_list_el = all('#progressBoxContent > ol > li')[9]
+    expect(registrar_progress_list_el).to have_content('Approved by Registrar - Done')
+
+    expect(page.find('#submissionApproved')).to have_content('Submission approved')
```
jcoyne: Have_selector with text:

spec/features/create_etd_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+
+    # check Argo for object
+    sleep(3) # waiting for Fedora/Solr
```
jcoyne: We use a pattern in other specs where we retry until the visit is successful. I think that will be more consistent than this.

spec/features/create_etd_spec.rb
ndushay: My point is that these tests in particular may require a longer wait time;  I can change it to "see spec_helper to increase this" sort of thing - would that suffice?

spec/features/create_etd_spec.rb
ndushay: I find these 3 puts statements super helpful, especially since we can't really track it all the way through ETD submission.   What's the harm?

I can certainly comment them out - can you explain your objection?

spec/features/create_etd_spec.rb
ndushay: yes, i can find a better way to express this (has_selector maybe)

spec/features/create_etd_spec.rb
ndushay: I will double-check -- this may have been an oversight; if not I will leave a better comment as to why we need it.

spec/features/create_etd_spec.rb
ndushay: I will check -- checkboxes are wonky because of how they are using javascript and images to show the checkmarks.

spec/features/create_etd_spec.rb
ndushay: true.  I will remove.  Doesn't seem like a big deal, tho.

spec/features/create_etd_spec.rb
ndushay: will remove

spec/features/create_etd_spec.rb
ndushay: I'll double check - it may be a link or a button or some weirdo javascript nasty.

spec/features/create_etd_spec.rb
ndushay: good tip.

spec/features/create_etd_spec.rb
ndushay: I know this is janky;  I specifically chose variable names to help.  Wouldn't a method be outside the block, so the let statements initializing the constant aren't executed?  I guess I could try doing a let statement at the top of the block, right after the `let(:xml_from_registrar)` -- is that what you meant?

spec/features/create_etd_spec.rb
ndushay: will do

spec/features/create_etd_spec.rb
ndushay: cool.  Then we can wait for the embargoed text to turn up as well.

ndushay:
@jcoyne I think you forgot to say "Nice job!  An excellent start!"   This was a total PITA as hydra_etd seemed to use different html/javascript for each step.  Getting a little praise along with the suggestions never goes amiss.

spec/features/create_etd_spec.rb
ndushay: (it's also handy if you need to find an ETD on stage)

jcoyne: Not a big deal, I thought these were leftover from developing the script. Just the very minor concern that this will clutter the rspec output.  Should we have a log file instead?

spec/features/create_etd_spec.rb
jcoyne: yes.

spec/features/create_etd_spec.rb
jcoyne: Sure but we've already increased it way beyond the default: https://github.com/sul-dlss/infrastructure-integration-test/blob/master/spec/spec_helper.rb#L32

jcoyne:
@ndushay I'm sorry for not being more positive. I appreciate the feedback and will try to improve.

spec/features/create_etd_spec.rb
ndushay: whoa!  30 seconds!  ok - removing.

spec/features/create_etd_spec.rb
ndushay: I'll leave them in but comment them out.

spec/features/create_etd_spec.rb
ndushay: removed -- it was an oversight that I spotted it.

spec/features/create_etd_spec.rb
ndushay: just FYI - i didn't know how to do this to get the contents of an attribute ...

PR #39, opened 2020-03-25
create_etd_spec: remove unnec "page."

ndushay:
## Why was this change made?

incorporating one last suggestion from PR #37 comments

na


ndushay:
and yes, i tested this change and the test completes green.

PR #50, opened 2020-03-31
Test for metadata deposit via spreadsheet

### LINKED ISSUE #18: modsulator-app needs integration tests
- test(s) to mimic the main uses of the modsulator-rails-app (via ... Argo??)

jermnelson:
## Why was this change made?
Fixes #18 

n/a


ndushay:
Jeremy - I don't profess to fully understand modsulator, but ... once it is loaded ... isn't it supposed to get into the descriptive metadata?  I thought that was the whole point.   

This test object doesn't seem to show that the result of running modsulator is that the data in the spreadsheet actually makes its way into the object in some way.  

Unfortuantely, you'll have to ask Andrew or Arcadia or someone to explain this ...

Issue #19, opened 2020-02-24
media objects (small files) need integration tests


ndushay:
e.g. movie and/or audio files via how we generally do it -- Andrew and/or Geoff W. can help us decide what the most common / most useful paths to test are.

ndushay:
I checked with Geoff;  
- media pretty much always starts with preassembly, and it tends to always go through in the same fashion

#### Sample Audio Manifest

```
source_id,filename,label,sequence,publish,preserve,shelve,resource_type
,gp921dm6826_a_pm.wav,Audio file 1,1,no,yes,no,audio
,gp921dm6826_a_sh.wav,Audio file 1,1,no,yes,no,audio
,gp921dm6826_a_sl.m4a,Audio file 1,1,yes,yes,yes,audio
,gp921dm6826_b_pm.wav,Audio file 2,2,no,yes,no,audio
,gp921dm6826_b_sh.wav,Audio file 2,2,no,yes,no,audio
,gp921dm6826_b_sl.m4a,Audio file 2,2,yes,yes,yes,audio
,gp921dm6826_img_01.tif,Image of media (1 of 6),3,no,yes,no,image
,gp921dm6826_img_02.tif,Image of media (2 of 6),4,no,yes,no,image
,gp921dm6826_img_03.tif,Image of media (3 of 6),5,no,yes,no,image
,gp921dm6826_img_04.tif,Image of media (4 of 6),6,no,yes,no,image
,gp921dm6826_img_05.tif,Image of media (5 of 6),7,no,yes,no,image
,gp921dm6826_img_06.tif,Image of media (6 of 6),8,no,yes,no,image```

#### Sample Video Manifest

````
source_id,filename,label,sequence,publish,preserve,shelve,resource_type
,qr373jp2115_1_p1_sh.mpeg,Video file 1,1,no,yes,no,video
,qr373jp2115_1_p1_sl.mp4,Video file 1,1,yes,yes,yes,video
,qr373jp2115_1_p1_thumb.jp2,Video file 1,1,yes,yes,yes,video
,qr373jp2115_1_p2_sh.mpeg,Video file 2,2,no,yes,no,video
,qr373jp2115_1_p2_sl.mp4,Video file 2,2,yes,yes,yes,video
,qr373jp2115_1_p2_thumb.jp2,Video file 2,2,yes,yes,yes,video
,qr373jp2115_1_p3_sh.mpeg,Video file 3,3,no,yes,no,video
[...]

PR #51, opened 2020-03-31
create_audio_object spec added

ndushay:
## Why was this change made?

The preassembly and accessioning of media objects is a bit of a special snowflake in our processing, mostly on the preassembly side.   This test creates a brand new audio object (with tiny files) and affirms that it makes it all the way through accessioning cleanly.

This is part of #19

Note that I intend to do a slight reorganization of this code when I add the media test -- I will run both the audio and the video objects through registration, preassembly and accessioning together as the longest wait times won't increase much by putting the two objects together.

na


ndushay:
I'm happy to take suggestions, but want to do them in a subsequent PR.  This code works now.  And I expect to be revising it to add a video object, in any case.

spec/features/create_audio_object_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+  # rubocop:disable Layout/IndentationWidth
+  let(:preassembly_manifest_csv) do
+# no leading spaces
+<<~CSV
+druid,object
+#{audio_druid},#{audio_druid}
+CSV
+  end
+  let(:media_manifest_csv) do
+# no leading spaces
```
jcoyne, line 36: Can you explain why?  Doesn't the squiggly heredoc strip leading spaces?

.rubocop_todo.yml
```diff
@@ -1,16 +1,24 @@
...
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
+# Offense count: 2
+# Cop supports --auto-correct.
+# Configuration parameters: EnforcedStyle.
+# SupportedStyles: squiggly, active_support, powerpack, unindent
+Layout/IndentHeredoc:
```
mjgiarlo, line 13: Since this cop claims to support auto-correct, are you open to running auto-correct and then re-generating the TODO config?

spec/features/create_audio_object_spec.rb
```diff
@@ -0,0 +1,156 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Create a new Audio object via Pre-assembly', type: :feature do
+  audio_druid = '' # used for preassembly manifest files
```
mjgiarlo, line 4: Seeing a variable declaration in an RSpec `describe` block throws me a bit. It's perfectly fine Ruby, of course, and RSpec is just Ruby. I'm wondering if perhaps `let!(:audio_druid) { '' }` would work just as well and might be less likely to cause me to blink?

spec/features/create_audio_object_spec.rb
mjgiarlo: I believe it ought to. This would be a nice, small change to make.

spec/features/create_audio_object_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+  let(:preassembly_manifest_csv) do
+# no leading spaces
+<<~CSV
+druid,object
+#{audio_druid},#{audio_druid}
+CSV
+  end
+  let(:media_manifest_csv) do
+# no leading spaces
+<<~CSV
```
mjgiarlo, line 37: Same re: above.

spec/features/create_audio_object_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+    expect(page).to have_selector('#document-contents-section > .resource-list > li.resource', text: 'Resource (2) text')
+    expect(page).to have_selector('#document-contents-section > .resource-list > li', text: 'Transcript')
+    audio_files.each do |af|
+      fname = druid_prefixed_name(af, audio_druid)
+      expect(page).to have_text(fname)
+    end
+  end
+end
+
+def druid_prefixed_name(old_filename, druid)
```
mjgiarlo, line 154: I'm up in the air on whether this method is a useful addition or more noise than it's worth. `"#{druid}_#{old_filename}"` is shorter than `druid_prefixed_name(old_filename, druid)` and it just as clear re: intent, I think? What do you think?

spec/features/create_audio_object_spec.rb
ndushay: sweet!  I breezed by this.

spec/features/create_audio_object_spec.rb
ndushay: I will try this, but I believe the memoized value messes things up;  I had a similar problem with ETD test.

ndushay: confirmed that `let` and `let!` will not work.

mjgiarlo: :hankey: 

thanks for giving it a shot!

spec/features/create_audio_object_spec.rb
ndushay: excellent point.  Will do.

ndushay:
@mjgiarlo @jcoyne 
- fixed the heredoc indentation
- killed the druid_prefixed_name method
- empirically proved that both `let` and `let!` memoize values and thus cannot be used for an unknown value going into `let` memoized heredoc.

PR #66, opened 2020-04-10
Test for Modsulator Changes to Druids

jermnelson:
## Why was this change made?
To confirm that the modsulator job successfully changed the titles in spreadsheet for two druids.

n/a


spec/features/deposit_via_xlsx_spec.rb
```diff
@@ -1,13 +1,31 @@
 # frozen_string_literal: true
 
+require 'rubyXL'
+require 'rubyXL/convenience_methods/cell'
```
jcoyne, line 4: why do we need this require?

spec/features/deposit_via_xlsx_spec.rb
```diff
@@ -1,13 +1,31 @@
...
+
 RSpec.describe 'Use Argo to upload metadata in a spreadsheet', type: :feature do
-  let(:start_url) { 'https://argo-stage.stanford.edu/view/druid:qc410yz8746' }
+  let(:argo_home) { 'https://argo-stage.stanford.edu/' }
+  let(:start_url) { "#{argo_home}view/druid:qc410yz8746" }
+  let(:druid1_url) { "#{argo_home}view/fh141gk9610" }
```
jcoyne, line 9: How are these druids created?

spec/features/deposit_via_xlsx_spec.rb
```diff
@@ -1,13 +1,31 @@
...
+  let(:title1) { RandomWord.phrases.next }
+  let(:title2) { RandomWord.phrases.next }
   let(:note) { RandomWord.phrases.next }
 
   before do
+    filled_xlsx = RubyXL::Parser.parse(spec_location)
+    sheet_one = filled_xlsx.worksheets[0]
+    sheet_one[2][3].change_contents title1
+    sheet_one[3][3].change_contents title2
+    filled_xlsx.write(spec_location)
```
jcoyne, line 21: how about writing to a new location rather than overwriting? Then we can cleanup just by removing the tempfile and we don't have to call out to git?

spec/features/deposit_via_xlsx_spec.rb
jermnelson: Removed `require` from latest commit.

spec/features/deposit_via_xlsx_spec.rb
jermnelson: These druids were manually created to match spreadsheet fixture.

spec/features/deposit_via_xlsx_spec.rb
jermnelson: Done, now writing to temp file.

spec/features/deposit_via_xlsx_spec.rb
jcoyne: Is it possible that we register new druids just for this test?  This will greatly reduce the overhead of setting up this test script when we clear out the staging environment.
If we can't we need to document how they need to be made so that we can initialize these fixtures so that the tests will work.

spec/spec_helper.rb
```diff
@@ -5,6 +5,8 @@
 require 'io/console'
 require 'pry-byebug'
 require 'random_word'
+require 'rubyXL'
+require 'rubyXL/convenience_methods/cell'
```
jcoyne, line 9: why do we need this?

jermnelson: The `rubyXL` gem is  the Excel spreadsheet

spec/features/deposit_via_xlsx_spec.rb
jermnelson: Done. Added new spec helper module to register the new druids.

Gemfile
```diff
@@ -13,5 +13,6 @@ gem 'random-word'
 gem 'rspec'
 gem 'rubocop', '~> 0.73.0'
 gem 'rubocop-rspec'
+gem 'rubyXL'
```
ndushay, line 16: could you add a comment that this is for updating Excel spreadsheets?

jermnelson: Done, just pushed up a comment in the gem file.

spec/spec_helper.rb
jcoyne: I see why we need rubyXL, but why this second require?

jcoyne: I see that this is for `change_contents` in XlsxHelpers.  Can we move that into that module and add a comment about that?

jermnelson: Done. Moved the `rubyXL/convenience_methods/cell` require to the xlsx_helpers with a comment.

PR #71, opened 2020-04-20
Extract APO to a constant

jcoyne:
## Why was this change made?

This will make it easier to reset the test suite if we ever clear stage.

no


spec/features/bulk_update_metadata_spec.rb
```diff
@@ -2,21 +2,22 @@
 
 RSpec.describe 'Use Argo to upload metadata in a spreadsheet', type: :feature do
   let(:argo_home) { 'https://argo-stage.stanford.edu/' }
-  let(:start_url) { "#{argo_home}view/druid:qc410yz8746" }
+  let(:start_url) { "#{argo_home}items/register" }
```
jcoyne, line 5: This prevents an unnecessary page load.

PR #87, opened 2020-05-05
make the default rake task running all the tests with rubocop

peetucket:
## Why was this change made?

Default rake task runs the test to match other projects; but travis should still only run rubocop

yes

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,4 +3,10 @@
 require 'rubocop/rake_task'
 RuboCop::RakeTask.new
 
-task default: [:rubocop]
+begin
+  require 'rspec/core/rake_task'
+  RSpec::Core::RakeTask.new(:spec)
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -3,4 +3,10 @@
...
 RuboCop::RakeTask.new
 
-task default: [:rubocop]
+begin
+  require 'rspec/core/rake_task'
+  RSpec::Core::RakeTask.new(:spec)
+rescue LoadError
+end
+
+task default: [:rubocop, :spec]
```
jcoyne, line 12: This doesn't make sense to me.  Rubocop checks this repo.  But `spec` tests everything outside of this repo.

peetucket: not sure ?  spec runs all the tests I think... travis only does rubocop

PR #93, opened 2020-05-18
Fix to bulk_update_metadata spec.

justinlittman:
## Why was this change made?
So that spec passes correctly.

## How was this change tested?
Tested against stage master.

## Which documentation and/or configurations were updated?
No

### DIFF
--- a/spec/features/bulk_update_metadata_spec.rb
+++ b/spec/features/bulk_update_metadata_spec.rb
@@ -43,8 +43,11 @@ RSpec.describe 'Use Argo to upload metadata in a spreadsheet', type: :feature do
       next unless tds[3].text == note
 
       tds[9].find('form > button').click
-      # Confirm delete in the popup
-      click_link 'Delete'
+      within('#confirm-delete-modal') do
+        # Confirm delete in the popup
+        click_link 'Delete'
+      end
+
       break
     end
     expect(page).to have_content "Bulk job for APO (#{APO}) deleted."


spec/features/bulk_update_metadata_spec.rb
```diff
@@ -43,8 +43,11 @@
       next unless tds[3].text == note
 
       tds[9].find('form > button').click
-      # Confirm delete in the popup
-      click_link 'Delete'
+      within('#confirm-delete-modal') do
```
jcoyne, line 46: Can you explain why we need this qualifier?

justinlittman: It didn't work before; it works now. I can't offer a satisfying explanation.

Wondering if it has something to do with the recent changes to modals?

jcoyne:
@justinlittman I was able to run spec/features/bulk_update_metadata_spec.rb  without this change and it passes.

justinlittman:
@jcoyne With argo master deployed?

jcoyne:
@justinlittman yes.

mjgiarlo:
FWIW, this test is passing for me w/o this change.

spec/features/bulk_update_metadata_spec.rb
peetucket: possibly another delete button was added elsewhere on the page and the simple 'Delete' selector became ambiguous?

jcoyne: @peetucket Capybara doesn't allow ambiguous selectors.

mjgiarlo:
@justinlittman can you try running what's in `master` now (with ffx, not chrome)?

justinlittman:
Master failed. Trying this branch.

justinlittman:
Both failed.

justinlittman:
Oh, if I sit and stare at it, master passes.

PR #98, opened 2020-05-18
Remove questions that don't make sense in this repo.

jcoyne:
## Why was this change made?
This repo is a test.  Typically we don't test our tests.

## How was this change tested?

It wasn't

## Which documentation and/or configurations were updated?
none

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,12 +1,2 @@
 ## Why was this change made?
 
-
-
-## How was this change tested?
-
-
-
-## Which documentation and/or configurations were updated?
-
-
-
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,2 +1,5 @@
 ## Why was this change made?
 
+## Was README.md updated if necessary?
+
+## Are there any configuration changes for shared_configs?


.github/pull_request_template.md
```diff
@@ -1,12 +1,2 @@
 ## Why was this change made?
```
justinlittman, line 2: The question about docs/configs still makes sense.

jcoyne: All the documentation is in the repo.  So asking "which" is odd.  Clearly if it's in the diff, then it's been updated. right?

For the configuration question should we be specific and say "shared configs"?  That is the only configuration outside of this repository.

justinlittman: From my perspective, the goal is to remind the developer and reviewer to write/review doc and config changes. That is accomplished by the words "documentation" and "configuration". Other than that, not too concerned with the wording.

.github/pull_request_template.md
```diff
@@ -1,12 +1,2 @@
...
 
-
-
-## How was this change tested?
-
-
-
-## Which documentation and/or configurations were updated?
```
jcoyne, line 9:
```suggestion
## Was README.md updated

## Are there any changes required in shared_configs
```

Issue #103, opened 2020-05-26
ensure all integration tests pass, at least when run individually


ndushay:
1.  Are there integration tests that always fail, even if run individually?

If so, they need to be rewritten to pass.
- is there a way to write the test so it passes more reliably?
- is the failing part of the test actually testing integration across services?  
    - If it is not testing cross service integration, can the failing part of the integration test be put in the relevant code base's test suite, and removed from the integration test?
- is there utility in having the test suite run, possibly with a "skip" at the point in the test when they always fail ... so we can at least execute the part that works?

From infradev meeting May 26, 2020 https://docs.google.com/document/d/1gckL1V0Iw6Vih7FO_-UXMJqbd97ReGCan2jbaLvEg-E/edit?skip_itp2_check=true&pli=1#heading=h.wkahme1uqz51

jcoyne:
I have not found that there are any that always fail.

mjgiarlo:
https://github.com/sul-dlss/infrastructure-integration-test/pull/104 moves things along but there remain two problems:

* The create_etd_spec using chrome consistently fails because the disseration upload step never completes. Not sure why.
* The edit_item_tags_spec using chrome fails in different spots. Again, not sure why.

Spent a couple hours chasing down different solutions to both of the above, to no avail. Things we can do to proceed:

1. Remove the chrome driver (:see_no_evil:)
1. Have more :eyes: and :brain: look at these specific failures
1. Do nothing and document the known failures
1. ???

@sul-dlss/infrastructure-team your thoughts?

ndushay:
https://rubydoc.info/github/jnicklas/capybara/master/Capybara%2FNode%2FActions:attach_file - does it need to somehow use a `visible: all` setting?

mjgiarlo:
@ndushay :speech_balloon: 
> [rubydoc.info/github/jnicklas/capybara/master/Capybara%2FNode%2FActions:attach_file](https://rubydoc.info/github/jnicklas/capybara/master/Capybara%2FNode%2FActions:attach_file) - does it need to somehow use a `visible: all` setting? 

Not sure that is "a thing" for `attach_file` (see also what I say below about `make_visible`). Also, we're using `visible: :all` already here: https://github.com/sul-dlss/infrastructure-integration-test/blob/master/spec/features/create_etd_spec.rb#L112

> Oh - and at one point, I was doing a refresh to get the file to show up when I was writing this spec, otherwise it seemed to load forever.

[...]

PR #104, opened 2020-05-27
Fix bugs and consolidate code in helpers

mjgiarlo:
Connects to #103 

## Why was this change made?

This commit fixes two known bugs:

* Using SdrClient::Deposit.run without handling the background job result ID that is returned. This broke the bulk_update_metadata spec.
* Changing the Argo registration URL (was removed in Argo). This broke two other specs.

As part of this work, pull related code into helpers so it is easier to vary behavior in the future without touching lots of files.

Not necessary.

No.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -27,3 +27,7 @@ Metrics/BlockLength:
     - it
     - context
     - scenario
+
+Style/IfInsideElse:
+  Exclude:
+    - 'spec/support/page_helpers.rb'
--- a/spec/features/create_preassembly_image_spec.rb
+++ b/spec/features/create_preassembly_image_spec.rb
@@ -5,7 +5,7 @@
 RSpec.describe 'Create new image object via Pre-assembly', type: :feature do
   druid = '' # used for HEREDOC preassembly manifest files (can't be memoized)
 
-  let(:start_url) { 'https://argo-stage.stanford.edu/items/register' }
+  let(:start_url) { 'https://argo-stage.stanford.edu/registration' }
   let(:preassembly_bundle_dir) { '/dor/staging/integration-tests/image-test' }
   let(:remote_manifest_location) { "preassembly@sul-preassembly-stage:#{preassembly_bundle_dir}" }
   let(:local_manifest_location) { 'tmp/manifest.csv' }
@@ -68,13 +68,10 @@ RSpec.describe 'Create new image object via Pre-assembly', type: :feature do
     # go to job details page, download result
     first('td > a').click
     expect(page).to have_content preassembly_project_name
+
     # wait for preassembly background job to finish
-    Timeout.timeout(Settings.timeouts.workflow) do
-      loop do
-        page.evaluate_script('window.location.reload()')
-        break if page.has_link?('Download')
-      end
-    end
+    reload_page_until_timeout!(text: 'Download', as_link: true)
+
     click_link 'Download'
     wait_for_download
     yaml = YAML.load_file(download)
@@ -90,12 +87,8 @@ RSpec.describe 'Create new image object via Pre-assembly', type: :feature do
     expect(files.last.text). to eq 'File image.jp2 (image/jp2, 64.2 KB, publish/shelve)'
 
     # Wait for accessioningWF to finish
-    Timeout.timeout(Settings.timeouts.workflow) do
-      loop do
-        page.evaluate_script('window.location.reload()')
-        break if page.has_text?('v1 Accessioned')
-      end
-    end
+    reload_page_until_timeout!(text: 'v1 Accessioned')
+
     expect(page).to have_selector('.blacklight-content_type_ssim', text: 'image') # filled in by accessioning
   end
 end
--- a/spec/features/reaccession_via_preassembly_spec.rb
+++ b/spec/features/reaccession_via_preassembly_spec.rb
@@ -38,13 +38,7 @@ RSpec.describe 'Reaccession from preassembly', type: :feature do
 
     first('td > a').click # Click to the job details page
 
-    # Wait for the background job to finish:
-    Timeout.timeout(Settings.timeouts.workflow) do
-      loop do
-        page.evaluate_script('window.location.reload()')
-        break if page.has_link?('Download')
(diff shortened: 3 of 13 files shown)


spec/features/create_preassembly_image_spec.rb
```diff
@@ -68,13 +68,10 @@
...
     expect(page).to have_content preassembly_project_name
+
     # wait for preassembly background job to finish
-    Timeout.timeout(Settings.timeouts.workflow) do
-      loop do
-        page.evaluate_script('window.location.reload()')
-        break if page.has_link?('Download')
-      end
-    end
+    reload_page_until_timeout!(text: 'Download')
```
ndushay, line 73: note that `.has_link?` is not the same as `text:`

spec/features/reaccession_via_preassembly_spec.rb
```diff
@@ -38,13 +38,7 @@
...
     first('td > a').click # Click to the job details page
 
-    # Wait for the background job to finish:
-    Timeout.timeout(Settings.timeouts.workflow) do
-      loop do
-        page.evaluate_script('window.location.reload()')
-        break if page.has_link?('Download')
-      end
-    end
+    reload_page_until_timeout!(text: 'Download')
```
ndushay, line 41: note that `.has_link?` is not the same as `text:`

spec/features/sdr_deposit_spec.rb
```diff
@@ -11,40 +11,23 @@
...
-    object_druid = nil
-    Timeout.timeout(Settings.timeouts.workflow) do
-      loop do
-        result = SdrClient::BackgroundJobResults.show(url: API_URL, job_id: job_id)
-        raise result[:output][:errors] if result[:output][:errors].present?
-
-        object_druid = result[:output][:druid]
-        break if object_druid
-      end
-    end
```
ndushay, line 37: I understand this is related to the bug you mentioned ... but don't we need to ensure it worked without error?  Or are we just waiting for 'v1 Accessioned` as sufficient?

spec/support/page_helpers.rb
```diff
@@ -0,0 +1,12 @@
...
+
+module PageHelpers
+  def reload_page_until_timeout!(text:)
+    Timeout.timeout(Settings.timeouts.workflow) do
+      loop do
+        page.evaluate_script('location.reload(true);')
+        break if page.has_text?(text)
+      end
+    end
+  end
```
ndushay, line 11: I think you should maybe have `reload_page_until_text_or_timeout` and `reload_page_until_link_or_timeout` ...

spec/features/create_preassembly_image_spec.rb
jcoyne: @ndushay I believe link text matches on `has_text` too.  I'm not sure that it being a link is critical to the purpose of this test.

spec/features/sdr_deposit_spec.rb
mjgiarlo: @ndushay This code just moved into `spec/support/deposit_helpers.rb`

mjgiarlo:
@jcoyne @ndushay thanks for the reviews! This is ready for another look.

spec/support/page_helpers.rb
```diff
@@ -0,0 +1,21 @@
...
+    Timeout.timeout(Settings.timeouts.workflow) do
+      loop do
+        # NOTE: Using passing `true` to this JavaScript function should force
+        #       the browser to bypass its cache.
+        page.evaluate_script('location.reload(true);')
+
+        # NOTE: This could have been a ternary but I was concerned about its
+        #       readability.
+        if as_link
+          break if page.has_link?(text)
```
jcoyne, line 14:
```suggestion
        break if (as_link && page.has_link?(text)) || page.has_text?(text)
```

spec/support/page_helpers.rb
```diff
@@ -0,0 +1,21 @@
...
+        else
+          break if page.has_text?(text)
+        end
```
jcoyne, line 17:
```suggestion
        break if (as_link && page.has_link?(text)) || page.has_text?(text)
```

.rubocop.yml
```diff
@@ -27,3 +27,7 @@ Metrics/BlockLength:
     - it
     - context
     - scenario
+
+Style/IfInsideElse:
+  Exclude:
+    - 'spec/support/page_helpers.rb'
```
jcoyne, line 33:
```suggestion
```

spec/support/page_helpers.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module PageHelpers
+  def reload_page_until_timeout!(text:, as_link: false)
```
jcoyne, line 4: This is great. 👍

spec/support/page_helpers.rb
mjgiarlo: @jcoyne I'm open to this but I wonder if it's much better than the ternary. It's essentially a ternary with different syntax, no? @sul-dlss/infrastructure-team ?

ndushay: I find the if / else easier to puzzle out at a glance, but it is only a slight preference.

jcoyne: How about `break if condition_satisfied?(text, as_link)`

mjgiarlo: @jcoyne I considered that too, but it felt gross to be passing all that state around. I'm open to that as well. I'm basically open to anything at least two potential approvers can agree on. :laughing:

justinlittman: `if / else` is clearer without mental exertion.

ndushay: i would say the code as it stands is an improvement over the pre-PR code, so it has my approval now or with any other way to handle this very tiny style concern.  I don't care - let's get the fix in!

mjgiarlo: Thanks, team. @jcoyne given :point_up:, are you open to changing the 'request changes' on this PR?

Issue #128, opened 2020-07-21
get create_etd_spec running in qa environment


ndushay:
We want the qa environment set up such that:

- [x] infrastructure-integration-tests
  - [x] [enable switching between QA and staging envs](https://github.com/sul-dlss/infrastructure-integration-test/pull/129)
- [x] ETD app
  - [x] [shared_configs](https://github.com/sul-dlss/shared_configs/tree/sul-hydra-etd-qa)
  - [x] [puppet](https://github.com/sul-dlss/operations-tasks/issues/2307)
  - [x] [capistrano](https://github.com/sul-dlss/hydra_etd/pull/661)
  - [x] [integration test works through ETD app steps](https://github.com/sul-dlss/infrastructure-integration-test/issues/128)
- [x] Argo
  - [x] [shared_configs](https://github.com/sul-dlss/shared_configs/tree/argo-qa)
    - [x] [regenerate DSA token to fix 401 errors](https://github.com/sul-dlss/shared_configs/commit/e2b1a48bcb9f0e04872022369805aa13acf3a74c)
  - [x] [puppet node-a](https://github.com/sul-dlss/puppet/blob/production/hieradata/node/argo-qa-a.stanford.edu.eyaml) and [puppet node-b](https://github.com/sul-dlss/puppet/blob/production/hieradata/node/argo-qa-b.stanford.edu.eyaml)
  - [x] [capistrano](https://github.com/sul-dlss/argo/blob/master/config/deploy/qa.rb)
  - [x] create [ETD APO in qa](https://argo-qa.stanford.edu/view/druid:bx911tp9024) based on [ETD APO in stage](https://argo-stage.stanford.edu/view/druid:bx911tp9024)
  - [x] [fix shibboleth in QA](https://github.com/sul-dlss/operations-tasks/issues/2314) - will know this is fixed once you can click e.g. [`etdSubmitWF`](https://argo-qa.stanford.edu/view/druid:wv456ww0683)
  - [x] [integration test works through end](https://github.com/sul-dlss/infrastructure-integration-test/issues/128)

mjgiarlo:
@ndushay :speech_balloon: 
> Additionally:
> 
> * [ ]  want dev to be able to run an etd from creation through accessioning, per [sul-dlss/hydra_etd#596 (comment)](https://github.com/sul-dlss/hydra_etd/issues/596#issue-662148601)
> 
[...]

PR #129, opened 2020-07-22
Allow running tests in both staging and QA environments

mjgiarlo:
Connects to #128

I can confirm that all the tests passed for me (after a couple runs, per usual).

## Why was this change made?

To allow running tests in staging and QA per #128.

Yes.

Not shared_configs, no, but local configs have been modified and documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # SDR Integration Tests
 
-This script drives a browser to do inter-system integration testing of SDR in the staging environment.
+This script drives a browser to do inter-system integration testing of SDR in the staging or QA environment.
 
 ## Installation
 
@@ -17,11 +17,11 @@ This script depends on having Firefox or Chrome downloaded.
 
 1. `bundle install`
 1. `rake webdrivers:chromedriver:update`
-1. Set `browser.driver` to `chrome` in `settings.local.yml`
+1. Set `browser.driver` to `chrome` in `config/settings.local.yml`
 
 ### Browser Window Size
 
-If you find you need to modify the default window size for either browser, copy the default settings for `browser.height` and `browser.width` from `settings.yml` to `settings.local.yml` and modify them to meet your needs.
+If you find you need to modify the default window size for either browser, copy the default settings for `browser.height` and `browser.width` from `config/settings.yml` to `config/settings.local.yml` and modify them to meet your needs.
 
 ## Prerequisites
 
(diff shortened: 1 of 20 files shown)


config/settings/qa.yml
```diff
@@ -0,0 +1,7 @@
+argo_url: 'https://argo-qa.stanford.edu'
+etd_url: 'https://etd-qa.stanford.edu'
+hydrus_url: 'https://sdr-qa.stanford.edu'
+preassembly_bundle_directory: '/dor/staging/integration-tests/image-test'
```
ndushay, line 4: we use the same directory?  Is there a downside/do we have to?

spec/features/bulk_update_metadata_spec.rb
```diff
@@ -6,34 +6,34 @@
...
     choose 'Spreadsheet input; load into objects'
     fill_in '3. Note', with: note
     click_button 'Submit'
 
     reload_page_until_timeout!(text: note)
 
-    # Delete's job run
+    # Delete job run
     job_rows = page.find_all('div#bulk-upload-table > table > tbody > tr')
-    job_rows.drop(1).each do |row|
```
ndushay, line 36: why was this dropping a row?  Why don't we need to any more?

README.md
```diff
@@ -37,13 +37,23 @@ Host *.stanford.edu
 
 ### SUNet Credentials
 
-If you tire of typing in your SUNet credentials over and over, you may add them to `settings.local.yml` (ignored by git). Copy the dummy values from `settings.yml` to get started. Do *not* add this file to version control, if you do this!
+If you tire of typing in your SUNet credentials over and over, you may add them to `config/settings.local.yml` (ignored by git). Copy the dummy values from `config/settings.yml` to get started. Do *not* add this file to version control, if you do this!
 
 ### ETD Credentials
 
-In order to run the tests in `spec/features/create_etd_spec.rb`, you will need to specify credentials required to authenticate connections to the ETD application. In order to do this, copy `settings.yml` to `settings.local.yml` and set the ETD username and password to the [values expected in our staging environment](https://github.com/sul-dlss/shared_configs/blob/a90c636b968a1ede4886a61dadc799dd5d162fe1/config/settings/production.yml#L34-L35).
+In order to run the tests in `spec/features/create_etd_spec.rb`, you will need to specify credentials required to authenticate connections to the ETD application.
```
ndushay, line 44: Add ...   "Set the username and password to the values in shared_configs for XXX and YYY [var names] in the appropriate environment, stage or qa"

README.md
```diff
@@ -2,7 +2,7 @@
 
 # SDR Integration Tests
 
-This script drives a browser to do inter-system integration testing of SDR in the staging environment.
+This script drives a browser to do inter-system integration testing of SDR in the staging or QA environment.
```
ndushay, line 5: where does it say "how to run in qa environment" ?   I assume just having the settings file isn't enough -- I would presumably have both of them at some point from doing my work ...

config/settings/qa.yml
mjgiarlo: I moved this into config just in case the value will vary between staging and QA.

README.md
mjgiarlo: In the Run Tests / SDR Environments section.

README.md
mjgiarlo: @ndushay if I understand, isn't that what's on lines 46-56 below?

spec/features/bulk_update_metadata_spec.rb
mjgiarlo: No need to drop a row since the check on line 38 is going to skip the header row anyway. It's noise, IMO.

config/settings/qa.yml
mjgiarlo: This moved from test setup to config just in case this value will vary between environments. The value is unchanged.

README.md
ndushay: ah - subheadings.  missed that.

README.md
ndushay: Does the fact that i missed it imply it should go higher?   Once I have my configs set up, I'm gonna want this info handier.

mjgiarlo: @ndushay how about I do a follow-up PR focused on re-organizing the README?

ndushay: sure - sounds good.  Not even that urgent.

PR #138, opened 2020-07-30
Etd should post a new ETD with the sunetid of the current user

### LINKED ISSUE #137: create etd spec needs to use a real sunetid to have permissions to fake posts from registrar

jcoyne:
This gives the current user the ability to edit this submission

## Why was this change made?

Fixes #137 (along with https://github.com/sul-dlss/hydra_etd/pull/730)

n/a


.rubocop_todo.yml
```diff
@@ -1,28 +1,28 @@
...
     - 'spec/support/authentication_helpers.rb'
 
-# Offense count: 91
+# Offense count: 101
 # Cop supports --auto-correct.
 # Configuration parameters: AutoCorrect, AllowHeredoc, AllowURI, URISchemes, IgnoreCopDirectives, IgnoredPatterns.
 # URISchemes: http, https
 Metrics/LineLength:
-  Max: 122
+  Max: 120
```
ndushay, line 28: I think 120 can be in `.rubocop.yml` but not worth holding up this PR.

jcoyne: I agree. This was just an autogenerate change though.

PR #148, opened 2020-08-31
Update dependencies

sul-devops-team:


.rubocop.yml
```diff
@@ -55,6 +82,9 @@ RSpec/ExampleLength:
 RSpec/MultipleExpectations:
   Enabled: false
 
+RSpec/MultipleMemoizedHelpers:
+  Enabled: false
```
jmartin-sul, line 86: turning this off seemed to be in the same spirit as us consciously turning off `RSpec/MultipleExpectations`.  didn't seem like a useful rule, considering our current testing patterns.

spec/support/config.rb
```diff
@@ -3,9 +3,10 @@
 require 'config'
 
 # NOTE: For some reason `File.expand_path(__dir__, '../..')` did not do the right thing.
-app_root = Pathname.new(__dir__).parent.parent + 'config'
+app_root = Pathname.new(__dir__).parent.parent
```
jmartin-sul, line 6: motivated by a complaint from a new cop about preferring interpolation over string concatenation.  this was actually not technically string concatenation, since the returned object is a `Pathname` (`+` is overridden on the `Pathname` class), so a bit of a false positive.  but taking a look at this and trying a switch to `Pathname#join` resulted in slightly clearer var naming while i was tweaking things, so seems like a win.

PR #170, opened 2020-11-03
Tests adding a Collection and Work in H2

jermnelson:
## Why was this change made?
Fixes https://github.com/sul-dlss/happy-heron/issues/203

n/a

n/a

### DIFF
--- a/spec/features/create_object_h2_spec.rb
+++ b/spec/features/create_object_h2_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe 'Use H2 to create an object', type: :feature do
   let(:collection_title) { RandomWord.nouns.next }
   let(:item_title) { "SUL Logo for #{collection_title}" }
   let(:start_url) { "#{Settings.h2_url}/dashboard" }
-  let(:user_email) { "#{AuthenticationHelpers.class_variable_get(:@@username)}@stanford.edu" }
+  let(:user_email) { "#{AuthenticationHelpers.username}@stanford.edu" }
 
   before do
     authenticate!(start_url: start_url,
@@ -23,7 +23,7 @@ RSpec.describe 'Use H2 to create an object', type: :feature do
     fill_in 'Contact email', with: user_email
 
     # Adds user to depositor field
-    fill_in 'Depositors', with: AuthenticationHelpers.class_variable_get(:@@username)
+    fill_in 'Depositors', with: AuthenticationHelpers.username
 
     click_button 'Deposit'
 
--- a/spec/features/create_object_h2_spec.rb
+++ b/spec/features/create_object_h2_spec.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Use H2 to create an object', type: :feature do
+  let(:collection_title) { RandomWord.nouns.next }
+  let(:item_title) { "SUL Logo for #{collection_title}" }
+  let(:start_url) { "#{Settings.h2_url}/dashboard" }
+  let(:user_email) { "#{AuthenticationHelpers.class_variable_get(:@@username)}@stanford.edu" }
+
+  before do
+    authenticate!(start_url: start_url,
+                  expected_text: 'Your collections')
+  end
+
+  scenario do
+    click_link '+ Create a new collection'
+
+    # Checks for specific content on create collection view
+    expect(page).to have_content('Manage release of deposits for discovery and download')
+
+    # Fills in basic collection information and saves
+    fill_in 'Collection name', with: collection_title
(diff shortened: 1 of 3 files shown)


spec/features/create_object_h2_spec.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Use H2 to create an object', type: :feature do
+  let(:collection_title) { RandomWord.nouns.next }
+  let(:item_title) { "SUL Logo for #{collection_title}" }
+  let(:start_url) { "#{Settings.h2_url}/dashboard" }
+  let(:user_email) { "#{AuthenticationHelpers.class_variable_get(:@@username)}@stanford.edu" }
```
jcoyne, line 7: Have you tried `AuthenticationHelpers.username` rather than `class_variable_get`

spec/features/create_object_h2_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    # Adds user to depositor field
+    fill_in 'Depositors', with: AuthenticationHelpers.class_variable_get(:@@username)
+
+    click_button 'Deposit'
+
+    expect(page).to have_content(collection_title)
+
+    # Deposit an item to the collection
+    find('header', text: collection_title).sibling('button').click
```
jcoyne, line 33: Can you do `click_button '+ Deposit to this collection', match: :first` instead?

spec/features/create_object_h2_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    fill_in 'Contact email', with: user_email
+    fill_in 'First name', with: 'Dana'
+    fill_in 'Last name', with: 'Scully'
+    fill_in 'Abstract', with: "An abstract for #{collection_title} logo"
+
+    fill_in 'Citation', with: 'All rights reserved'
+
+    find('input#work_agree_to_terms').click
+
+    find_button('Deposit').click
```
jcoyne, line 57: This triggers an asynchronous action.  How does it wait until the deposit is finished?

jcoyne (changes requested):

spec/features/create_object_h2_spec.rb
jermnelson: That works, also updated line 26 to use `AuthenticationHelpers.username`

spec/features/create_object_h2_spec.rb
jermnelson: Do we care that adding a deposit is in the same collection we created? On stage (for my login) there is a long list of collections (see below screenshot) and this was how I could ensure that the object being deposited was using the same collection that was just created.

Screen Shot 2020-11-03 at 4 13 55 PM

spec/features/create_object_h2_spec.rb
jermnelson: I believe the `expect` statement on line 60 waits until the deposit is finished.

spec/features/create_object_h2_spec.rb
jermnelson: Done

spec/features/create_object_h2_spec.rb
jcoyne: Ah, that's a good point.

spec/features/create_object_h2_spec.rb
jcoyne: Line 60 is a synchronous reponse though.  That will be printed out right away while a background job will transfer the file to SDR.

jermnelson: Added `reload_page_until_timeout` call that waits for the druid to appear on work status page.

jermnelson:
@jcoyne ready for re-review. Added `reload_page_until_timeout` to address async comment.

mjgiarlo:
@jermnelson Looks great, thank you! Have you run this against QA and stage? (I'm not sure it'd work in QA yet given it needing some shib attr changes, hence my curiosity.)

jermnelson:
> @jermnelson Looks great, thank you! Have you run this against QA and stage? (I'm not sure it'd work in QA yet given it needing some shib attr changes, hence my curiosity.)

I've been running this test on stage but I don't have the correct rights to run the test on H2 in QA.

mjgiarlo:
@jermnelson :speech_balloon: 
> 
> I've been running this test on stage but I don't have the correct rights to run the test on H2 in QA.

OK, great. Thank you.

mjgiarlo:
@jcoyne can you re-review?

jcoyne (approved):
Looks good, thank you.

PR #206, opened 2021-03-08
Fix for H2 spec.

justinlittman:
## Why was this change made?
Green is my favorite color.

NA

NA

### DIFF
--- a/spec/features/create_object_h2_spec.rb
+++ b/spec/features/create_object_h2_spec.rb
@@ -68,7 +68,7 @@ RSpec.describe 'Use H2 to create an object', type: :feature do
     expect(page).to have_content(item_title)
 
     # This happens asynchronously, it might take a bit
-    expect(page).to have_content('https://sul-purl-stage')
+    expect(page).to have_content('https://purl-stage.stanford.edu')
 
     # Opens Argo and searches on title
     visit Settings.argo_url


spec/features/create_object_h2_spec.rb
```diff
@@ -68,7 +68,7 @@
     expect(page).to have_content(item_title)
 
     # This happens asynchronously, it might take a bit
-    expect(page).to have_content('https://sul-purl-stage')
+    expect(page).to have_content('https://purl-stage.stanford.edu')
```
jcoyne, line 71: this does not work in the qa environment.

jcoyne (changes requested):

spec/features/create_object_h2_spec.rb
mjgiarlo: Sounds like we need this as a new settings variable so we can have env-specific variables.

jcoyne: I think we can configure h2-qa because I think they're both pointing at the same purl (but different c-names)

justinlittman: Done (https://github.com/sul-dlss/shared_configs/commit/707455d321af3f38c960cc3bcf47685dc4470fc5) and deployed.

jcoyne (approved):

PR #212, opened 2021-03-19
create an object after a regitering an apo to verify default rights

### LINKED ISSUE #209: Add an integration test for creating an AdminPolicy in Argo
In order to verify that default rights settings are appropriately applied when creating an AdminPolicy, an integration test should be added that:

- [x] Creates an APO with specified default rights fields
- [x] Registers an object with that APO
- [x] Verify that the object has those fields set

peetucket:
## Why was this change made?

Fixes #209 - update create apo test to also register an object with this APO and verify default rights show are shown


spec/features/create_apo_spec.rb
```diff
@@ -24,5 +28,29 @@
...
+    # wait for object to be registered
+    find('td[aria-describedby=data_status][title=success]')
+    object_druid = find('td[aria-describedby=data_druid]').text
+    # puts "object_druid: #{object_druid}" # useful for debugging
+
+    visit "#{Settings.argo_url}/view/#{object_druid}"
+
+    # wait for registrationWF to finish and verify default access rights
+    reload_page_until_timeout!(text: 'v1 Registered', with_reindex: true)
+    expect(page).to have_content "Access Rights:\nstanford"
```
jcoyne, line 54: Can we also check copyright, license and terms of use?

peetucket: sounds reasonable, let me give that a try

spec/features/create_apo_spec.rb
```diff
@@ -24,5 +34,34 @@
...
+    visit "#{Settings.argo_url}/view/#{object_druid}"
+
+    # wait for registrationWF to finish and verify default access rights
+    reload_page_until_timeout!(text: 'v1 Registered', with_reindex: true)
+    expect(page).to have_content "Access Rights:\n#{rights.downcase}"
+
+    # these are in the cocina model data, which is hidden by default
+    expect(page).to have_content(:all, terms_of_use)
+    expect(page).to have_content(:all, copyright)
+    #expect(page).to have_content(:all, license_code) # This should be there, but isn't in cocina!
```
peetucket, line 65: the license isn't in the cocina model for the item at all, which seems wrong

spec/features/create_apo_spec.rb
peetucket: added license, terms and copyright to APO to verify they propogate to the item associated with the APO, but oddly enough i'm not seeing the license code in the cocina model of the item at all, which this test is looking for

peetucket: We don't seem to be mapping the license into cocina here: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/from_fedora/access.rb#L18-L27

spec/features/create_apo_spec.rb
```diff
@@ -24,5 +34,34 @@
...
+    visit "#{Settings.argo_url}/view/#{object_druid}"
+
+    # wait for registrationWF to finish and verify default access rights
+    reload_page_until_timeout!(text: 'v1 Registered', with_reindex: true)
+    expect(page).to have_content "Access Rights:\n#{rights.downcase}"
+
+    # these are in the cocina model data, which is hidden by default
+    expect(page).to have_content(:all, terms_of_use)
+    expect(page).to have_content(:all, copyright)
+    # expect(page).to have_content(:all, license_code) # This should be there, but isn't in cocina!
```
jcoyne, line 65: If you deploy the main branch of DSA, then it does indeed show the license:
Screen Shot 2021-03-22 at 4 20 47 PM

peetucket: Alrighty then...that's a recent change.  Let me update the test to reflect the fact it will be the URI and not the code.  I'll wait for DSA to be deployed to -prod with the dependency updates.

ndushay: i think you could look for "by-nc" as included in the string, if you preferred but whatever.

jcoyne (approved):
Thank you. This will be super helpful.

PR #215, opened 2021-03-23
add temporary access_indexing spec that does NOT run by default

### LINKED ISSUE #214: add (temporary) test for cocina access rights indexing
Because we are simultaneously working on updating 
- mappings from rightsMetadata.xml to cocina
- indexing rights from cocina

and we want the index values to stay the same (i.e. when fedora xml is XXX and it is mapped to cocina, then indexed, then the index values from the cocina match the index values that would have come from fedora)

use an rspec tag and set it up so this test is NOT run by default.

Once we are using only cocina for rights editing and indexing, this test will have no value and we can remove it.

ndushay:
## Why was this change made?

To have an easy way to test the impact of the following on the Access Rights facet:
- cocina model access changes
- dor-services-app access mapping changes (from and to fedora)
- dor-indexing-app changes to use cocina model instead of fedora for indexing
- argo hook ins to cocina access as it affects indexing.

@andrewjbtw thought this was a good idea, btw.

tagging @mjgiarlo so he is aware of this.

Fixes #214 

.rspec was changed NOT to execute the new spec by default.

### DIFF
--- a/spec/features/access_indexing_spec.rb
+++ b/spec/features/access_indexing_spec.rb
@@ -1,0 +1,97 @@
+# frozen_string_literal: true
+
+# NOTE: this spec is intended for use testing changes to dor-services-app mappings to cocina and
+#  testing changes to dor_indexing_app to index access from cocina
+#  to run:  bundle exec rspec --tag indexing spec/features/access_indexing_spec.rb
+RSpec.describe 'Argo rights changes result in correct Access Rights facet value', type: :feature, indexing: true do
+  let(:random_word) { RandomWord.phrases.next }
+  let(:object_label) { "Object Label for #{random_word}" }
+  let(:start_url) { "#{Settings.argo_url}/registration" }
+  let(:source_id) { "access-rights-test:#{random_word}" }
+
+  before do
+    authenticate!(start_url: start_url,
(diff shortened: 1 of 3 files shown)


spec/features/access_indexing_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+  let(:source_id) { "access-rights-test:#{random_word}" }
+
+  before do
+    authenticate!(start_url: start_url,
+                  expected_text: 'Register DOR Items')
+  end
+
+  scenario do
+    # fill in registration form
+    select 'integration-testing', from: 'Admin Policy'
```
jcoyne, line 19: I think this breaks if we don't have this Admin Policy or there are two of them with this name.

ndushay: most of our integration tests use this admin policy, and in fact, I just copied this part from another test.

jcoyne: Ah, looks like many of our other tests have the same problem.

ndushay: so ... would you care to approve this PR?   asking for a friend.

jcoyne (approved):

Issue #41, opened 2020-03-25
test embargo release


ndushay:
can we do this?
- create an object with a very soon embargo release time (15 min?)
- ensure object is embargoed 
   - shows that way in Argo
   - is preserved
   - didn't get released to stacks inappropriately (shelve)
   - didn't get released to purl inappropriately (publish)
 - do something so object comes off embargo (can we change the embargo time, run the code that removes the embargo?)
 - ensure object is no longer embargoed (argo, stacks, purl ...)

jcoyne:
Embargo release typically happens once a day at 2:16am: https://github.com/sul-dlss/dor-services-app/blob/master/config/schedule.rb#L24

PR #217, opened 2021-03-24
add embargo_spec

ndushay:
## Why was this change made?

Partly because I'm rewriting the dor_indexing_app handling of embargo metadata (see sul-dlss/dor_indexing_app/issues/508) and want a way to test it in situ, and partly because I know it would be useful to have a test for embargo release (see issue #41).

I would be open to tagging this test so it's not always run by default, if that is preferred -- but I think it's a useful test to keep in the code base.


ndushay:
a key thing is that the embargo shows up in the purl correctly - is there any way to test this.

jcoyne:
I'm pretty sure that embargo metadata has never been sent to purl: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/publish/metadata_transfer_service.rb#L32-L42

andrewjbtw:
The embargo release date has to go into the rightsMetadata that gets published in the public XML so that Purl can display the release date. If that date is not stored in the Cocina access metadata, it has to be pulled from the embargo metadata for publishing.

jcoyne:
@andrewjbtw is this a new requirement?  I see rightsMetadata copied into the public xml, but I don't see where the release date is added: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/publish/public_xml_service.rb#L25

jcoyne:
I don't see that purl actually does anything with that node (https://github.com/sul-dlss/purl/search?q=embargoReleaseDate) .  @andrewjbtw do you have any examples where we see this on the purl?

andrewjbtw:
All of these display the release date at the top of the file listing box: https://argo.stanford.edu/catalog?utf8=%E2%9C%93&all=true&f%5Bcontent_type_ssim%5D%5B%5D=file&f%5Bprocessing_status_text_ssi%5D%5B%5D=Accessioned&f%5Brights_descriptions_ssim%5D%5B%5D=citation&f%5Bembargo_release_dtsim%5D%5B%5D=%5B2021-04-09T00%3A00%3A00.000Z+TO+2030-01-01T23%3A59%3A59.000Z%5D&commit=submit

I guess it's actually sul-embed that is using the data: https://github.com/sul-dlss/sul-embed/search?q=embargoReleaseDate

You can see it in the purl XML:

```xml
<access type="read">
  <machine>
    <embargoReleaseDate>2021-05-05T00:00:00Z</embargoReleaseDate>
    <none/>
  </machine>
</access>
```
I think that's currently being published straight from the rightsMetadata. Storing the value only once in Cocina (instead of the current twice in Fedora) seems like clearly the right way to go, but if/when we stop publishing the rightsMetadata as it appears in Fedora, we have to make sure that date still gets to the purl.

jcoyne:
So we can just click the "purl" link from argo and look for "Access is restricted until [date]":
Screen Shot 2021-04-07 at 10 53 30 PM

spec/features/embargo_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+    # check Argo for object (wait for embargo info)
+    Timeout.timeout(Settings.timeouts.workflow) do
+      loop do
+        visit "#{Settings.argo_url}/view/#{druid}"
+        break if page.has_text?("This item is embargoed until #{embargo_date.strftime('%F').tr('-', '.')}", wait: 1)
+      end
+    end
+
+    # check Argo facet field with 6 month embargo
+    visit "#{Settings.argo_url}/catalog?search_field=text&q=#{druid}"
```
jcoyne, line 125: what is `search_field=text` for?

ndushay:
- The embargo checking for hydrus COULD be added to the `deposit_via_hydrus` spec.
- The embargo checking for ETDs has been added to the create_etd spec.
- The embargo checking for H2 (and thus, cocina mapping) does NOT work, which the test confirmed, so that test is skipped by default via tags.

aaron-collier (approved):

PR #221, opened 2021-03-31
authentication allows for automatic duo pushes

ndushay:
## Why was this change made?

I was running an integration test and it was failing as soon as I approved in Duo.  Apparently some of our systems (or maybe all???) do automatic duo pushes now.

### DIFF
--- a/spec/support/authentication_helpers.rb
+++ b/spec/support/authentication_helpers.rb
@@ -7,7 +7,7 @@ module AuthenticationHelpers
     # View the specified starting URL
     visit start_url
 
-    submit_credentials
+    submit_credentials(expected_text)
 
     using_wait_time 100 do
       # Once we see this we know the log in succeeded.
@@ -15,7 +15,7 @@ module AuthenticationHelpers
     end
   end
 
-  def submit_credentials
+  def submit_credentials(expected_text)
     self.username ||= username_from_config_or_prompt
     self.password ||= password_from_config_or_prompt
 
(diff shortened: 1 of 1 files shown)


spec/support/authentication_helpers.rb
```diff
@@ -27,6 +27,8 @@ def submit_credentials
     sleep 1
     click_button 'Login'
 
+    return if find('Pushed a login request to your device')
```
ndushay, line 30: Note that this will keep looking until our timeout for the text, which will slow things down if the text is NOT present ... can someone be more clever about this?

jcoyne:
I wonder if UIT is rolling this out to only some people, because I haven't seen it. I still get sent to the page where it asks me what flavor of 2FA I want to use.

ndushay:
I have no idea, but I couldn't get the tests to run without this change, and it should allow all of us to run tests without a problem.

mjgiarlo (approved):

jcoyne:
I tested this an it no longer sends me a push.

mjgiarlo:
@jcoyne :speech_balloon: 
> This makes running the first test take 100 seconds longer.

Should I add to dev planning?

jcoyne:
yes

mjgiarlo:
@jcoyne done

ndushay:
Can the wait time be reduced to help?

I was unable to run integration tests without these changes, so we're going to have to figure out a way for this to work for diff cases that no single developer can test.  :-(

spec/support/authentication_helpers.rb
```diff
@@ -27,8 +27,17 @@ def submit_credentials
     sleep 1
     click_button 'Login'
 
-    within_frame('duo_iframe') do
-      click_button 'Send Me a Push'
+    # did we already get authenticated?
+    return if page.has_text?(expected_text, wait: 100)
```
ndushay, line 31: is this 100 seconds we're waiting?  I thought it was milliseconds

spec/support/authentication_helpers.rb
```diff
@@ -27,8 +27,17 @@ def submit_credentials
...
+
+    # did we already push, but not authenticated?
+    begin
+      return if page.has_text?('Pushed a login request to your device', wait: 100)
```
ndushay, line 35: another wait spot that may have too long a wait time.

mjgiarlo:
@ndushay :speech_balloon: 
> Can the wait time be reduced to help?
> 
> I was unable to run integration tests without these changes, so we're going to have to figure out a way for this to work for diff cases that no single developer can test. :-(

Flagged for dev planning (next week)

PR #232, opened 2021-04-13
fix embargo specs

ndushay:
## Why was this change made?

fixed the h2 embargo spec so it should pass when DSA is correct.

### DIFF
--- a/spec/features/embargo_h2_spec.rb
+++ b/spec/features/embargo_h2_spec.rb
@@ -12,10 +12,8 @@
 RSpec.describe 'H2 item created with embargo; change embargo in Argo',
                type: :feature, stage_only: true, embargo: true, skip_for_default: true do
   let(:collection_title) { RandomWord.nouns.next }
-  let(:item_title) { "Text file for #{collection_title}" }
+  let(:item_title) { "Text file #{RandomWord.nouns.next} in H2 collection" }
   let(:user_email) { "#{AuthenticationHelpers.username}@stanford.edu" }
-  let(:purl_xml_embargo_prefix) { '<access type="read">\s*<machine>\s*<embargoReleaseDate>' }
-  let(:purl_xml_embargo_suffix) { '</embargoReleaseDate>\s*<none.*>\s*</machine>\s*</access>' }
 
   before do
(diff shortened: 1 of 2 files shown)


spec/features/embargo_h2_spec.rb
```diff
@@ -12,10 +12,10 @@
 RSpec.describe 'H2 item created with embargo; change embargo in Argo',
                type: :feature, stage_only: true, embargo: true, skip_for_default: true do
   let(:collection_title) { RandomWord.nouns.next }
-  let(:item_title) { "Text file for #{collection_title}" }
```
ndushay, line 15: whoops - not unique.  hahaha

spec/features/embargo_h2_spec.rb
```diff
@@ -12,10 +12,10 @@
...
+  let(:item_title) { "Text file #{RandomWord.nouns.next} in H2 collection" }
   let(:user_email) { "#{AuthenticationHelpers.username}@stanford.edu" }
-  let(:purl_xml_embargo_prefix) { '<access type="read">\s*<machine>\s*<embargoReleaseDate>' }
-  let(:purl_xml_embargo_suffix) { '</embargoReleaseDate>\s*<none.*>\s*</machine>\s*</access>' }
+  let(:purl_xml_embargo_prefix) { '<access type="read">\s*<machine>\s*<none.*>\s*<embargoReleaseDate>' }
+  let(:purl_xml_embargo_suffix) { '</embargoReleaseDate>\s*</machine>\s*</access>' }
```
ndushay, line 18: I'm not in love with caring about whether the embargo element is first or last here.   Ideas?   I'm not looking to deal with the entire purl.xml

jcoyne: xpath query?

ndushay: so slurp up the entire purl and treat it as xml?   could do.

ndushay:
@jcoyne Now using xpath to find embargo date on purl page and ensure there is only one such node.   This passed on whatever is currently on stage as of right now.

jcoyne (approved):

PR #239, opened 2021-04-22
Remove the edit tags spec

jcoyne:
## Why was this change made?
This has been moved into https://github.com/sul-dlss/argo/pull/2703

Since this has no dependencies on any service outside of Argo's test suite, it's best to run it there since it will run on every CI run.


justinlittman (approved):
But this test would reliably pass ...

PR #243, opened 2021-04-23
Tweak to create_etd_spec.

justinlittman:
## Why was this change made?
Less bashing of head against desk.

NA

NA


jmartin-sul (approved):
sure, why not (not that we want to be adding `sleep` wherever, but i think we're all judicious about this sort of thing, and what's one more second in a long running integration test suite 😄)

PR #245, opened 2021-04-26
Check public XML using dor-services-app

### LINKED ISSUE #241: for embargo tests, check xml that would be sent to purl, rather than purl itself
Thus allowing h2 and hydrus embargo tests to verify the purl xml is correct without needing to go to purl itself, which isn't available in qa.

mjgiarlo:
Fixes #241

## Why was this change made?

Use the dor-services-app public XML endpoint instead of relying on a purl service to be operational, which we do not have in QA. Due to this change, and increasing general parity between QA and staging, we can move the H2 embargo testing into the other H2 test and we can get rid of some RSpec tags. Do this for faster, more reliable testing in both envs.

I was able to see both affected specs (Hydrus and H2) pass in both QA and stage.

No

No


mjgiarlo:
I'd like for @ndushay to review this PR before we merge it. Other reviews also welcome!

aaron-collier (approved):

config/settings/staging.yml
```diff
@@ -6,4 +6,5 @@ preassembly_bundle_directory: '/dor/staging/integration-tests/image-test'
 preassembly_host: 'sul-preassembly-stage'
 preassembly_url: 'https://sul-preassembly-stage.stanford.edu'
 sdrapi_url: 'https://sdr-api-stage.stanford.edu'
-purl_url: 'https://purl-stage.stanford.edu'
+dor_services:
```
ndushay, line 9: any reason to keep two levels here?   not a big deal, just curious.

spec/support/purl_helpers.rb
```diff
@@ -10,22 +11,10 @@ module PurlHelpers
...
-      loop do
-        visit "#{Settings.purl_url}/#{druid}.xml"
-        break if html.match?('rightsMetadata')
-
-        sleep 1
-      end
-    end
-
-    purl_ng_xml = Nokogiri::XML(html)
+    Dor::Services::Client.configure(url: Settings.dor_services.url, token: Settings.dor_services.token)
```
ndushay, line 15: should there be a comment for the module or the method that clarifies this is not looking at the purl, but is looking at the public_xml that will be sent to the purl?   Or maybe even a module rename?

ndushay (approved):
This looks splendid.   A few line comments, but nothing show stopping.

config/settings/staging.yml
mjgiarlo: @ndushay It's arbitrary, admittedly, but I did this to be consistent with the pattern we use everywhere else:

```yaml
dor_services:
  url: 'foo'
  token: 'bar'
```

spec/support/purl_helpers.rb
mjgiarlo: I'll add a comment and rename. thanks!

PR #247, opened 2021-04-27
Tweak to bulk_update_metadata spec to allow for slow upload.

justinlittman:
## Why was this change made?
Test was failing due to slow upload.

Why is the upload slow? No idea.

### DIFF
--- a/spec/features/bulk_update_metadata_spec.rb
+++ b/spec/features/bulk_update_metadata_spec.rb
@@ -28,6 +28,7 @@ RSpec.describe 'Use Argo to upload metadata in a spreadsheet', type: :feature do
     choose 'Spreadsheet input; load into objects'
     fill_in '3. Note', with: note
     click_button 'Submit'
+    expect(page).to have_content('Bulk processing started')
 
     reload_page_until_timeout!(text: note)
 


spec/features/bulk_update_metadata_spec.rb
```diff
@@ -28,6 +28,7 @@
     choose 'Spreadsheet input; load into objects'
     fill_in '3. Note', with: note
     click_button 'Submit'
+    sleep 5 # For some reason, somtimes this upload is slow.
```
jcoyne, line 31: typo 'sometimes'

justinlittman: No, nothing appears after the submit and the check below that involves a refresh interrupts the form submission. We're in some odd Capybara / browser twilight zone here. Open to suggestions.

And fixed typo.

jcoyne: How about we add this first? https://github.com/sul-dlss/argo/pull/2715

justinlittman: With that change, I was able to replace wait.

jcoyne (approved):

PR #264, opened 2021-05-20
check event service to confirm preservation replicated new versions to all cloud endpoints

### LINKED ISSUE #249: basic testing of preservation functionality
Augment `create_preassembly_image_spec.rb` to check the event log in Argo to confirm that both versions have been replicated to all preservation catalog cloud endpoints.

Requires the event logging to be implemented first: https://github.com/sul-dlss/preservation_catalog/issues/1685

Implementing this (and the supporting pres cat ticket) would eliminate some of the manual testing follow up required by e.g. sul-dlss/preservation_robots#310, sul-dlss/preservation_catalog#1589, and sul-dlss/preservation_catalog#1614.

jmartin-sul:
## Why was this change made?

so that infra integration test suite can confirm that preservation and replication are working, since it's already implicitly exercised by accessioning.

closes #249

also, adds a (hopefully) generally useful helper method and pattern for querying the event service.

~~leaving in draft, as accessioning was wedged on stage when i tried to test last night (`provenanceMetadata` wasn't getting created, but i did test just this change by commenting out all of this class except the new event service check, and then hardcoding a known druid and version from stage -- obviously not a realistic enough test to merge this change).~~

n/a

nope

### DIFF
--- a/spec/features/create_preassembly_image_spec.rb
+++ b/spec/features/create_preassembly_image_spec.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'druid-tools'
+
 # Preassembly requires that files to be included in an object must be available on a mounted drive
 # To this end, files have been placed on Settings.preassembly_host at Settings.preassembly_bundle_directory
 RSpec.describe 'Create and reaccession object via Pre-assembly', type: :feature do
@@ -129,8 +131,40 @@ RSpec.describe 'Create and reaccession object via Pre-assembly', type: :feature 
     yaml = YAML.load_file(download)
     expect(yaml[:status]).to eq 'success'
 
-    visit "#{Settings.argo_url}/view/druid:#{yaml[:pid]}"
+    prefixed_druid = "druid:#{yaml[:pid]}"
+    latest_version = version + 1
 
-    reload_page_until_timeout!(text: "v#{version + 1} Accessioned", with_reindex: true)
+    visit "#{Settings.argo_url}/view/#{prefixed_druid}"
+    reload_page_until_timeout!(text: "v#{latest_version} Accessioned", with_reindex: true)
+
+    # The below confirms that preservation replication is working: we only replicate a
+    # Moab version once it's been written successfully to on prem storage roots, and
+    # we only log an event to dor-services-app after a version has successfully replicated
+    # to a cloud endpoint.  So, confirming that both versions of our test object have
+    # replication events logged for all three cloud endpoints is a good basic test of the
+    # entire preservation flow.
+
+    druid_tree_str = DruidTools::Druid.new(prefixed_druid).tree.join('/')
+
+    latest_s3_key = "#{druid_tree_str}.v000#{latest_version}.zip"
+    reload_page_until_timeout!(text: latest_s3_key, with_events_expanded: true)
+
+    # the event log should eventually contain an event for replication of each version that
+    # this test created to every endpoint we archive to
+    poll_for_matching_events!(prefixed_druid) do |events|
+      (1..latest_version).all? do |cur_version|
+        cur_s3_key = "#{druid_tree_str}.v000#{cur_version}.zip"
+
+        %w[aws_s3_west_2 ibm_us_south aws_s3_east_1].all? do |endpoint_name|
+          events.any? do |event|
+            event[:event_type] == 'druid_version_replicated' &&
+              event[:data]['parts_info'] &&
+              event[:data]['parts_info'].size == 1 && # we only expect one part for this small object
+              event[:data]['parts_info'].first['s3_key'] == cur_s3_key &&
+              event[:data]['endpoint_name'] == endpoint_name
+          end
+        end
+      end
+    end
   end
 end
--- a/spec/support/event_helpers.rb
+++ b/spec/support/event_helpers.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module EventHelpers
+  # pass in a block that returns true if the event list has the desired event(s)
+  def poll_for_matching_events!(prefixed_druid)
+    Dor::Services::Client.configure(url: Settings.dor_services.url, token: Settings.dor_services.token)
+    object_client = Dor::Services::Client.object(prefixed_druid)
+
+    Timeout.timeout(Settings.timeouts.events.poll_for) do
+      loop do
+        events = object_client.events.list
+        break if yield(events)
+
+        sleep Settings.timeouts.events.poll_interval
+      end
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem 'capybara'
 gem 'capybara_table'
 gem 'config'
 gem 'dor-services-client', '~> 6.34'
+gem 'druid-tools' # for constructing druid tree strings easily
 gem 'faraday' # etds require POST request from registrar
 gem 'pry-byebug'
 gem 'rake'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -233,6 +233,7 @@ DEPENDENCIES
   capybara_table
   config
   dor-services-client (~> 6.34)
+  druid-tools
   faraday
   pry-byebug
   rake
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -2,6 +2,10 @@ timeouts:
   capybara: 60
   workflow: 300
   bulk_action: 200
+  events:
+    poll_for: 240
+    poll_interval: 2
+
 
 browser:
   driver: firefox
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -57,6 +57,7 @@ RSpec.configure do |config|
   config.include AuthenticationHelpers
   config.include DepositHelpers
   config.include DownloadHelpers
+  config.include EventHelpers
   config.include PageHelpers
   config.include PublicXmlHelpers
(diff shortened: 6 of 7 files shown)


jmartin-sul:
stage is unwedged, tested changes as they are in this PR, spec worked for me

spec/features/create_preassembly_image_spec.rb
```diff
@@ -129,8 +131,36 @@
     yaml = YAML.load_file(download)
     expect(yaml[:status]).to eq 'success'
 
-    visit "#{Settings.argo_url}/view/druid:#{yaml[:pid]}"
+    prefixed_druid = "druid:#{yaml[:pid]}"
```
ndushay, line 134: maybe add a comment before this section to indicate this is all about confirming preservation replication to could endpoints is working

ndushay (changes requested):
My main questions with this:
1.  will it work on qa (do we have full preservation qa?)  Could we have prescat qa replicate to local storage dir so it would work?
2. could there be a loop to check if prescat replication is finished, so it waits less than 30 seconds when replication happens faster?   That's a lot to add to a test, and it's just a guess, correct?
3.  maybe we could discuss at infradev what the best way to do this is ... not sure ...

ndushay:
I also wonder if we really want to check this for stage and qa, or if we'd like a check for it for prod.  Maybe we can update a version of a test object in prod that isn't published, and have a cron test this ... weekly???

jmartin-sul:
> My main questions with this:
> 
>     1. will it work on qa (do we have full preservation qa?)  Could we have prescat qa replicate to local storage dir so it would work?
> 

Good call!  I'll admit I only tested this against stage.  Luckily, based on testing this morning, it seems that this should on QA, and seems that pres cat is fully wired up there (though in confirming this, I did stumble on some unexpected behavior, for which I can comment separately):

https://argo-qa.stanford.edu/view/druid:dh325ky8841
Screen Shot 2021-05-24 at 6 30 50 PM
Screen Shot 2021-05-24 at 10 57 07 AM
Screen Shot 2021-05-24 at 10 57 23 AM

Additionally I double-checked manually using this script from the wiki:
https://github.com/sul-dlss/preservation_catalog/wiki/Investigating-a-druid-with-replication-errors

And both versions showed as present on all 3 cloud endpoints.  


>     2. could there be a loop to check if prescat replication is finished, so it waits less than 30 seconds when replication happens faster?   That's a lot to add to a test, and it's just a guess, correct?
> 

Sure.  The `sleep` that's there is, tbh, totally based on gut feeling from watching pres cat workers replicate stuff in the past.

My guess is that it usually takes about 30 seconds to zip a version and ship it off to all 3 endpoints and then mark it as replicated to each endpoint (at which point a message is sent to DSA, see https://github.com/sul-dlss/preservation_catalog/pull/1687).  E.g., I had a number of seconds to capture each of the above screenshots, even with the worker pool being otherwise totally idle at the time.

My goal here was just to keep the test from hammering DSA (and constantly refershing Argo) when I knew it probably wouldn't have an answer for a little bit.  Maybe not worth worrying about though? 🤷 I was on the fence there.

If it is worth worrying about, what about an alternate approach that just starts polling immediately, but sleeps for a couple seconds each loop iteration?  That might have the benefit of making it easier to copycat the pattern -- presumably other tests that might be interested in whether an event has made it to the event service might expect a bit of lag, if only for the occasional slow network request to DSA itself.

I don't personally see a substantial difference in terms of test wait time between polling pres cat and polling DSA, since I would expect that last request to DSA to take at most a second or two on most runs.  The benefit of polling DSA only is that it means one less client/token/service to add to this test suite, and no new REST methods to add to pres cat (or no need to reach under the hood and run a rails runner script on a pres box, depending on how you'd poll pres).

FWIW my first instinct when it came to adding an integration test for preservation/replication was to query pres cat directly in some way, but emitting events to DSA and then querying DSA ended up feeling both easier to write a test for and more generally useful (since this is a good event to capture if replication related debugging is needed later).

>     3. maybe we could discuss at infradev what the best way to do this is ... not sure ...

Sure, happy to discuss further, I'm not in a huge rush to get this merged, though I am still in favor of the general approach taken here.

jmartin-sul:
> I also wonder if we really want to check this for stage and qa, or if we'd like a check for it for prod. Maybe we can update a version of a test object in prod that isn't published, and have a cron test this ... weekly???

I do like the idea of some sort of ongoing prod health check along these lines:  to my knowledge, we currently check our cloud endpoints to make sure that everything we expect to be there is in fact still present.  But we don't yet check to see what portion of our locally preserved objects are as-yet unreplicated to the cloud (because we were wary of noise since we knew pres cat would spin up with a huge replication backlog, and then we just never circled back on that audit question -- maybe worth checking to see if there's a ticket).

But this is something that at least I'm personally in favor of testing on stage or qa, in the same way that we regression test any other major area of functionality when assessing things like weekly dependency updates or code changes that tweak how services interoperate.  To my knowledge, there's no existing end-to-end test of anything preservation related in a deployed prod like environment, the way there is with for example accessioning.  Also useful as a basic smoke test for changes like https://github.com/sul-dlss/preservation_robots/pull/311

jmartin-sul:
went ahead and tried the easy/obvious fix for the tiny bug i stumbled on:  https://github.com/sul-dlss/dor-services-app/pull/2858

spec/features/create_preassembly_image_spec.rb
```diff
@@ -129,8 +131,43 @@
...
+
+    # The below confirms that preservation replication is working: we only replicate a
+    # Moab version once it's been written successfully to on prem storage roots, and
+    # we only log an event to dor-services-app after a version has successfully replicated
+    # to a cloud endpoint.  So, confirming that both versions of our test object have
+    # replication events logged for all three cloud endpoints is a good basic test of the
+    # entire preservation flow.
+
+    # hacky, give pres cat replication pipeline a chance to finish
+    sleep 30
```
ndushay, line 148: I would prefer this is reduced in case we can get the test to finish faster;  maybe make a separate timeout value for preservation on the order of one minute for the reload_until??

spec/features/create_preassembly_image_spec.rb
```diff
@@ -129,8 +131,43 @@
...
+
+    druid_tree_str = DruidTools::Druid.new(prefixed_druid).tree.join('/')
+
+    latest_s3_key = "#{druid_tree_str}.v000#{latest_version}.zip"
+    reload_page_until_timeout!(text: latest_s3_key, with_events_expanded: true)
+
+    # the event log should eventually contain an event for replication of each version that
+    # this test created to every endpoint we archive to
+    poll_for_matching_events!(prefixed_druid) do |events|
```
ndushay, line 157: I wonder if it makes sense for the helper method to be closer to "poll_for_event_matching(druid, version, something)".   Or not.  Just a thought, certainly not a blocker.

spec/support/event_helpers.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module EventHelpers
+  # pass in a block that returns true if the event list has the desired event(s)
+  def poll_for_matching_events!(prefixed_druid)
+    Dor::Services::Client.configure(url: Settings.dor_services.url, token: Settings.dor_services.token)
+    object_client = Dor::Services::Client.object(prefixed_druid)
+
+    Timeout.timeout(Settings.timeouts.workflow) do
```
ndushay, line 9: This is where the preservation timeout could be used; 

also, this could possibly become "poll_for_matching_replication_event(druid, version, s3key)"

ndushay (changes requested):
One must do about the timeout;  I would say a separate s3 timeout and s3 checking interval (could be 2s? more?);  one food for thought thing.

Also, should I be concerned that https://argo.stanford.edu/view/druid:bb480hh7974 shows this:

image

Did the event service start getting preservation replication events later than Feb?

ndushay:
here's one accessioned today:

https://argo.stanford.edu/view/druid:bc219cf6344

image

jcoyne:
@ndushay I think you might be confusing the events service with the events datastream.   They are two different things. The events service is in a database and the intent is that it replaces the events datastream (since events is not part of cocina).

ndushay:
@jcoyne you are right - I didn't click the right thing in Argo - d'oh!

So John:

1.  please do something about the 30 seconds.   maybe 3 config settings:  pause before starting to check, how long to keep checking, and checking interval (I'm okay with 2s or even more)

2.  consider a refactor of the event_helper, but if you prefer it how you have it, I can live with it, no problem.

spec/features/create_preassembly_image_spec.rb
jmartin-sul: since i added a delay to the polling loop, and since my worry was mostly about sending way too many requests to an API that'll respond quickly, i just got rid of this `sleep` entirely.  there's now a 2 second delay between each attempt to poll the event service.

spec/features/create_preassembly_image_spec.rb
jmartin-sul: i somewhat prefer it as is, because it allows the polling logic to be re-used easily, while keeping the logic about what events are relevant in the specific test that's invoking the polling.  though i can also see the appeal of hiding away the logic for which events match.

spec/support/event_helpers.rb
jmartin-sul: specific timeout values added

jmartin-sul:
fixed up version still consistently passes against stage

ndushay (approved):

PR #266, opened 2021-05-24
Improve 2 tests, clarify settings needs

ndushay:
## Why was this change made?

52ffab5  - clarify the README and the settings.yml files re: token needed for both dor-services-app and dor-services-client
2d180df - fixes a race condition in the create_object_no_files test for less flappiness (yay!)
4133a17 - h2 test uses settings h2_purl_url so it works for both qa and stage.

yes, it was.

### DIFF
--- a/spec/features/create_object_h2_spec.rb
+++ b/spec/features/create_object_h2_spec.rb
@@ -71,7 +71,7 @@ RSpec.describe 'Use H2 to create an object', type: :feature do
     # Checks if title is on resulting display
     expect(page).to have_content(item_title)
     # This happens asynchronously, it might take a bit
-    expect(page).to have_content('https://sul-purl-stage.stanford.edu')
+    expect(page).to have_content(Settings.h2_purl_url)
 
     # Opens Argo and searches on title
     visit Settings.argo_url
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ If you are experiencing timeout errors when running tests, you may override the 
 
 ### Set Dor-Services-App Tokens
 
-Some integration tests use the `dor-services-client` to interact with the `dor-services-app`. In order to successfully use the dor-services-app API, you must first have a token set. (To generate dor-services-app tokens, see the [dor-services-app README](https://github.com/sul-dlss/dor-services-app#authentication).) Note that you'll need to do this for each environment (currently: stage and QA). Place the value in `config/settings/{ENV}.local.yml`. See `config/settings.yml` for the expected YAML syntax.
+Some integration tests use the `dor-services-client` to interact with the `dor-services-app`, others directly (or indirectly) use the `dor-services-app`.  In order to successfully use the dor-services-app API, you must first have a token set. (To generate dor-services-app tokens, see the [dor-services-app README](https://github.com/sul-dlss/dor-services-app#authentication).) Note that you'll need to do this for each environment (currently: staging and qa). Place the value in `config/settings/{ENV}.local.yml`. See `config/settings.yml` for the expected YAML syntax.
 
 ### Set SUNet Credentials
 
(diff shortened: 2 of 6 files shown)


spec/features/create_object_h2_spec.rb
```diff
@@ -71,7 +71,7 @@
     # Checks if title is on resulting display
     expect(page).to have_content(item_title)
     # This happens asynchronously, it might take a bit
-    expect(page).to have_content('https://sul-purl-stage.stanford.edu')
+    expect(page).to have_content(/https:\/\/(sul-|)purl-(stage|qa)\.stanford\.edu/)
```
jcoyne, line 74: Should this be a config variable rather than a regex?

spec/features/create_object_no_files_spec.rb
```diff
@@ -35,6 +35,7 @@
     reload_page_until_timeout!(text: 'v1 Registered', with_reindex: true)
 
     # add accessionWF
+    expect(page).to have_text('Successfully updated index for') # ensure we see this first
```
jcoyne, line 38: Isn't seeing "v1 Registered" on line 35 enough to indicate we are on the show page?  Why do we need to also look for this?

spec/features/create_object_h2_spec.rb
ndushay: I think it would need to be what is in h2 shared_configs, which may change - kind of a config of a config.  The above works now, and right now qa is set to stage ... so it should be future-proof?

spec/features/create_object_no_files_spec.rb
ndushay: because it's a race condition:  if the wf is added BEFORE the "Successfully updated index" message, then we don't see the "added AccessionWF" message ever.  I was hitting this over and over yesterday.

spec/features/create_object_h2_spec.rb
jcoyne: I guess what I'm saying is `expect(page).to have_content Settings.purl_link` is easier to understand

spec/features/create_object_no_files_spec.rb
jcoyne: I'm having trouble understanding the race condition here.  Can you explain it to me?  Line 35 should loop over reindexing N times until `v1 Registered` is on the page.  Only then will it add the workflow, so, how is it racing?

spec/features/create_object_h2_spec.rb
ndushay: ok, added it to settings and clarified where it was coming from with comments.   I understood you the first time but was concerned the value in qa.yml  would be puzzling to get right.

spec/features/create_object_no_files_spec.rb
ndushay: If v1 is accessioning, but the 'Successfully updated index for' message doesn't appear before we add the workflow, then the workflow is added, but the message 'Added accessionWF' may never display because it is stomped over by 'Successfully updated index' message.   I saw this happen a couple times:   the test hung waiting for 'Added accessionWF' message but I could see 'Successfully updated index' as the message and the accessioning WF had started.

ndushay:
back to you, @jcoyne

spec/features/create_object_no_files_spec.rb
jcoyne: Why is "Successfully updated index for" appearing?  That shouldn't happen once `v1 Registered` appears.  This sounds like a bug in `reload_page_until_timeout!`

jcoyne: Maybe the check for "Successfully updated index" should go here: https://github.com/sul-dlss/infrastructure-integration-test/blob/216d6d675c1ec4bed5ae384e93f924c407944057/spec/support/page_helpers.rb#L14

ndushay: good idea - i'll do that.

jcoyne (approved):

PR #296, opened 2021-08-30
when re-indexing, still force a page refresh

peetucket:
## Why was this change made?

The `spec/features/sdr_deposit_spec.rb` started failing because the page in Argo never shows the object as completed accessioning, even though it had (as could be verified in another browser window).  This just forces the page to refresh even if a re-index was requested.


aaron-collier:
This fixed all of my failures. Merging.

aaron-collier (approved):

PR #304, opened 2021-09-08
Fix h2 spec

ndushay:
## Why was this change made?

h2 spec never succeeded due to introduced error

### DIFF
--- a/spec/features/create_object_h2_spec.rb
+++ b/spec/features/create_object_h2_spec.rb
@@ -6,18 +6,19 @@ RSpec.describe 'Use H2 to create an object', type: :feature do
   let(:user_email) { "#{AuthenticationHelpers.username}@stanford.edu" }
 
   before do
-    authenticate!(start_url: "#{Settings.h2_url}/dashboard", expected_text: 'Create a new collection')
+    authenticate!(start_url: "#{Settings.h2_url}/dashboard", expected_text: 'Dashboard')
   end
 
   scenario do
+    # remove modal for deposit in progress, if present, waiting a bit for some rendering
     click_button 'No' if page.has_content?('Continue your deposit', wait: Settings.post_authentication_text_timeout)
 
     # CREATE COLLECTION
     click_link '+ Create a new collection'
-    # Checks for specific content on create collection view
+    # Checks for specific content in create collection view
     expect(page).to have_content('Manage release of deposits for discovery and download')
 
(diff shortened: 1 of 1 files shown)


spec/features/create_object_h2_spec.rb
```diff
@@ -35,11 +38,11 @@
     expect(page).to have_content(collection_title)
     expect(page).to have_content('+ Deposit to this collection')
 
-    item_id = page.current_url.split('/').last
+    collection_id = page.current_url.split('/').last
```
ndushay, line 41: this was confusing me for a bit

spec/features/create_object_h2_spec.rb
```diff
@@ -59,7 +62,6 @@
     click_button 'Continue'
 
     # Work Deposit view
-    fill_in 'What\'s changing?', with: 'Initial metadata'
```
ndushay, line 62: this was a bug introduced -- the PR for reserve a purl was old, so I guess this changed.

spec/features/create_object_h2_spec.rb
```diff
@@ -6,17 +6,20 @@
...
   end
 
   scenario do
+    # remove modal for deposit in progress, if present
     click_button 'No' if page.has_content?('Continue your deposit', wait: Settings.post_authentication_text_timeout)
-    click_link '+ Create a new collection'
 
-    # Checks for specific content on create collection view
+    # CREATE COLLECTION
+    expect(page).to have_content('+ Create a new collection')
```
jcoyne, line 17: This is unnecessary duplication. Is there a reason for it?

spec/features/create_object_h2_spec.rb
jcoyne: Yeah, this is no longer shown on the first version

jcoyne (changes requested):

spec/features/create_object_h2_spec.rb
ndushay: i can try without;  there was a lot of debugging going on with what came before.

ndushay: yup - it's fine without.  w00t!  Maybe it's more robust now.  I hope so.

ndushay:
@jcoyne ready for re-review.  Thanks for the suggestion.

jcoyne (approved):

PR #328, opened 2022-01-11
create_object_no_files: fix modal selector.

lwrubel:
## Why was this change made?
create_object_no_files test was broken. Runs on staging now.

### DIFF
--- a/spec/features/create_object_no_files_spec.rb
+++ b/spec/features/create_object_no_files_spec.rb
@@ -45,7 +45,7 @@ RSpec.describe 'Use Argo to create an object without any files', type: :feature 
 
     # open a new version
     click_link 'Open for modification'
-    within '.modal-dialog' do
+    within '.modal-body' do
       select 'Admin', from: 'Type'
       fill_in 'Version description', with: 'opening version for integration testing'
       click_button 'Open Version'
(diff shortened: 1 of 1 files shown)


spec/features/create_object_no_files_spec.rb
```diff
@@ -45,7 +45,7 @@
 
     # open a new version
     click_link 'Open for modification'
-    within '.modal-dialog' do
+    within '.modal-body' do
```
jcoyne, line 48: I'm pretty sure the class we're looking for is `modal-dialog`. See https://github.com/sul-dlss/argo/blob/c3fef087978e09fe80770ce3af66538a505e704a/app/views/catalog/show.html.erb#L4

jcoyne (changes requested):

spec/features/create_object_no_files_spec.rb
jcoyne: The code on argo staging is currently in flux and there is a bug in the code. The test was looking correctly for modal-dialog.

lwrubel: Got it. Should I just close this PR for now?

jcoyne: Yes, I think so

PR #346, opened 2022-02-14
Virtual objects integration test

### LINKED ISSUE #112: Add a spec for creating and viewing a virtual object
Virtual objects are serialized in a different way by dor-services-app, so this will test that argo is able to deserialize.

lwrubel:
## Why was this change made?
Resolves #112 to create a virtual object with multiple constituents via Bulk Action. 

[Example virtual object with 2 constituents](https://argo-stage.stanford.edu/view/druid:hy591qz5434). (3:59 to run)
[Example virtual object with 11 constituents](https://argo-stage.stanford.edu/view/druid:mr596mj1416). (14:08 to run)

Is 14 minutes a reasonable amount of time to have a test take? The number of constituents is set in a let definition--I can see how long it takes with other numbers of constituents if that's useful. 

No further set-up or configuration needed for this test.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,6 @@ gem 'rubocop', '~> 1.24'
 gem 'rubocop-rake'
 gem 'rubocop-rspec', '~> 2.1'
 gem 'rubyXL' # for updating Excel spreadsheets
+gem 'sdr-client', '~> 0.67'
 gem 'selenium-webdriver'
-gem 'sdr-client', '~> 0.58'
 gem 'webdrivers', '~> 4.0'
--- a/spec/features/create_virtual_object_spec.rb
+++ b/spec/features/create_virtual_object_spec.rb
@@ -5,7 +5,7 @@ RSpec.describe 'Use Argo to create a virtual object with constituent objects', t
   let(:filename_group) { ['example.tiff', 'example.jp2'] }
   let(:csv_path) { File.join(DownloadHelpers::PATH, 'virtual-object.csv') }
   let(:virtual_objects_description) { RandomWord.phrases.next }
-  let(:num_constituents) { 11 }
+  let(:num_constituents) { Settings.number_of_constituents }
 
   before do
     authenticate!(start_url: start_url, expected_text: 'Welcome to Argo!')
@@ -26,9 +26,6 @@ RSpec.describe 'Use Argo to create a virtual object with constituent objects', t
 
     # Create parent object
     parent_druid = deposit_object
-
-    # Avoid occasional errors with preservation workflow
-    sleep 2
 
     # Create CSV: parent_druid, object_druid, object_druid
     virtual_object_row = object_druids.clone
@@ -69,7 +66,7 @@ RSpec.describe 'Use Argo to create a virtual object with constituent objects', t
       end
     end
 
-    visit "#{start_url}/view/#{parent_druid}?beta=true"
+    visit "#{start_url}/view/#{parent_druid}"
     reload_page_until_timeout!(text: 'v2 Accessioned', with_reindex: true)
 
     # Confirm constituent druids are listed in Content
--- a/spec/features/create_virtual_object_spec.rb
+++ b/spec/features/create_virtual_object_spec.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Use Argo to create a virtual object with constituent objects', type: :feature do
+  let(:start_url) { Settings.argo_url }
+  let(:filename_group) { ['example.tiff', 'example.jp2'] }
+  let(:csv_path) { File.join(DownloadHelpers::PATH, 'virtual-object.csv') }
+  let(:virtual_objects_description) { RandomWord.phrases.next }
+  let(:num_constituents) { 11 }
+
+  before do
+    authenticate!(start_url: start_url, expected_text: 'Welcome to Argo!')
+  end
+
+  scenario do
+    ensure_token
+
+    # Create constituent objects
+    object_druids = []
+
+    num_constituents.times do
+      object_druid = deposit_object(filenames: filename_group)
+      object_druids << object_druid
+    end
+
+    puts object_druids
+
+    # Create parent object
+    parent_druid = deposit_object
+
+    # Avoid occasional errors with preservation workflow
+    sleep 2
+
+    # Create CSV: parent_druid, object_druid, object_druid
+    virtual_object_row = object_druids.clone
+    virtual_object_row.unshift(parent_druid)
+
+    CSV.open(csv_path, 'w') do |csv|
+      csv << virtual_object_row
+    end
+
+    # Use Bulk Actions to upload CSV
+    click_link 'Bulk Edits'
+    click_link 'Bulk Actions (asynchronous)'
+    expect(page).to have_content 'Bulk Actions'
+    click_link 'New Bulk Action'
+    expect(page).to have_content 'New Bulk Action'
+    select 'Create virtual object(s)', from: 'bulk_action_action_type'
+    expect(page).to have_content 'Create one or more virtual objects'
+    find('input#bulk_action_create_virtual_objects_csv_file').attach_file(csv_path)
+    find('textarea#bulk_action_description').fill_in(with: virtual_objects_description)
+    click_button 'Submit'
+
+    expect(page).to have_content 'Bulk action was successfully created.'
+
+    Timeout.timeout(Settings.timeouts.bulk_action) do
+      loop do
+        page.driver.browser.navigate.refresh
+
+        relevant_bulk_action = find(:xpath, "//tr[td = '#{virtual_objects_description}']")
+        within(relevant_bulk_action) do
+          status_text = all('td')[3].text
+          next unless status_text == 'Completed'
+
+          results_text = all('td')[4].text
+          expect(results_text).to eq('1 / 1 / 0')
+        end
+
+        break
+      end
+    end
+
+    visit "#{start_url}/view/#{parent_druid}?beta=true"
+    reload_page_until_timeout!(text: 'v2 Accessioned', with_reindex: true)
+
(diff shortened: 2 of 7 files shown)


lwrubel:
This test currently creates two child objects, each with its own set of tiff/jp2 files. @andrewjbtw recommended that we test with more than ten child objects. I'm wondering it, in that case, the files all need to be different for each object? Could I just re-use the same image files with each child object, or might that reduce how valid a test this is?

andrewjbtw:
It could be run with the same image files in each druid. Since the druids are going to be unique, I think it will be possible to identity issues despite the files being the same.

lwrubel:
Do I have the SDR Client version specified correctly in the Gemfile? It needs to be at least 0.67.1.

Gemfile
```diff
@@ -19,6 +19,6 @@ gem 'rubocop', '~> 1.24'
 gem 'rubocop-rake'
 gem 'rubocop-rspec', '~> 2.1'
 gem 'rubyXL' # for updating Excel spreadsheets
+gem 'sdr-client', '~> 0.67.1'
```
justinlittman, line 22: I would try `'~> 0.67` and then `bundle update` and see what version you get in `Gemfile.lock`.

lwrubel: Since you get `0.67.1` with bundle update, I've changed this back to `'~> 0.67'`, thanks.

spec/features/create_virtual_object_spec.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Use Argo to create a virtual object with constituent objects', type: :feature do
+  let(:start_url) { Settings.argo_url }
+  let(:filename_group) { ['example.tiff', 'example.jp2'] }
+  let(:csv_path) { File.join(DownloadHelpers::PATH, 'virtual-object.csv') }
+  let(:virtual_objects_description) { RandomWord.phrases.next }
+  let(:num_constituents) { 11 }
```
jcoyne, line 8: Does the number have to be so high to test this feature?  Would 2 constituents be enough?  It's valuable to have tests, but having tests that run fast is important too or people will not want to use this tool.

spec/features/create_virtual_object_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      object_druid = deposit_object(filenames: filename_group)
+      object_druids << object_druid
+    end
+
+    puts object_druids
+
+    # Create parent object
+    parent_druid = deposit_object
+
+    # Avoid occasional errors with preservation workflow
```
jcoyne, line 30: Can you be more specific about the errors? We don't want to have sleep anywhere, so if you think we need it, at least document what error you are avoiding.

jcoyne (changes requested):

spec/features/create_virtual_object_spec.rb
justinlittman: This number was requested by Andrew, but @lwrubel was going to perform some time testing to determine if it is realistic.

jcoyne (commented):
The fixtures here are big and will bloat the git repo. Can you use smaller fixtures like https://github.com/samvera/hyrax/blob/main/spec/fixtures/image.jp2 ?

spec/features/create_virtual_object_spec.rb
jcoyne: 14 minutes is how much time it takes to run the rest of the suite, so I'd much prefer we use 2 if that's okay.

lwrubel:
Thanks for the link to a smaller jp2, @jcoyne. I created a tiff from that and it helps the whole test run faster, as well. 

It still takes ~7 minutes to run with 11 constituents. I'll change it back to 2, unless @andrewjbtw still has a concern about that.

jcoyne:
@lwrubel another possibility is to throw that disputed number into `conf/settings.yml` (as `number_of_constituents: 2`) and then someone wanting to run with more could do `SETTINGS__NUMBER_OF_CONSTITUENTS=11 bundle exec rspec`

spec/features/create_virtual_object_spec.rb
andrewjbtw: I think it would be useful to have an automated test that takes longer than the usual integration test but doesn't have to be run all the time. I would not test dedicated changes to virtual objects by using only two constituents, so if this test only uses two it would not take the place of testing that I do manually.

lwrubel: Thanks for that context, @andrewjbtw. Would the approach of running it in a way that's configurable, such as `SETTINGS__NUMBER_OF_CONSTITUENTS=11 bundle exec rspec` like @jcoyne suggested work? @mjgiarlo you were also interested in this test, I think, so I'm wondering if this would also work for what you were envisioning.

andrewjbtw: Yes - configurable would be great.

mjgiarlo: @lwrubel hi! any number of constituents would suffice for me.

spec/features/create_virtual_object_spec.rb
lwrubel: I tried taking the sleep out just to see if I was being overly cautious and I got one of the errors again just now: 
`Error: sdr-ingest-transfer : problem with sdr-ingest-transfer (BackgroundJob: 233047): {:errors=>[{:title=>"Preservation error", :detail=>"Version mismatch in /dor/workspace/zh/212/bg/7530/zh212bg7530/metadata/versionMetadata.xml, expected 1, found 2"}]} `

I've also seen an error that says, `Error: validate-bag : Bag validation failure(s) for druid:yj420my5239: [{:required_file_not_found=>"Deposit bag required file /services-disk-stage/store2/deposit/yj420my5239/data/metadata/versionMetadata.xml not found"}] `

I'll test some more with sleep back in, and update the comment to describe the error as an "validate-bag error or sdr-ingest-transfer preservation error with versionMetadata.xml", if that sounds informative enough. If there's wording that would better describe this, let me know and I can put that in.

jcoyne: I'm not sure how the sleep would prevent such an error. That error took place asynchronously in the accessionWF/preservationWF robots.

jcoyne: @mjgiarlo is it possible that Laura has found a bug in the `Create virtual object` where it's not checking to see if all the constituents are accessioned before messing with them?

mjgiarlo: @jcoyne @lwrubel we have been seeing version-related errors when I deploy the `cocinize-reset-content#3331` branch of DSA, which is currently in QA (not stage). And I am not sure I've figured out how to deal with that bug yet. Were the above errors in QA? (And I assume we've not seen them in stage?)

mjgiarlo: @jcoyne when validating constituents, a number of checks are made, including whether each constituent is either Opened or Accessioned: https://github.com/sul-dlss/dor-services-app/pull/3379/files#diff-65455a7c13e2140bbb2d58b75073ad8d010e5da90bca66b52483f30e1e492507R29-R33 so we are making this check

mjgiarlo: Are the versionMetadata errors on the virtual object side, the constituent side, or both?

lwrubel: I've been seeing them on stage, after the constituent and parent objects have been created and the bulk action run. When the virtual object is being created (a v2 being accessioned), the error happens. Here's an example: https://argo-stage.stanford.edu/view/druid:zh212bg7530

mjgiarlo: @lwrubel Oh, interesting. I'm curious if you see them in QA too or if that runs clean.

lwrubel:
@jcoyne, while I've been able to make a number_of_constituents a setting and use it, I haven't figured out how to pass it in the command-line as you're describing so that it overrides the settings.yml value. Is there an example of this somewhere?

spec/features/create_virtual_object_spec.rb
lwrubel: @mjgiarlo: just got it on QA: https://argo-qa.stanford.edu/view/druid:jn028rp2502

lwrubel: Also am seeing this error on QA, "Creating some or all virtual objects failed because some objects are not combinable: Problematic objects for druid:mx887np6058: undefined method `externalIdentifier' for #<Event:0x0000000006ec2510>".

mjgiarlo: @lwrubel Well, that's odd. It looks like an event object is being included as a constituent!

mjgiarlo: This was due to my DSA branch assuming VersionService.close returned a Cocina object. It returns an event instead.

spec/features/create_virtual_object_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+          results_text = all('td')[4].text
+          expect(results_text).to eq('1 / 1 / 0')
+        end
+
+        break
+      end
+    end
+
+    visit "#{start_url}/view/#{parent_druid}?beta=true"
```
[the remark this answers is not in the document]
lwrubel: Have removed the beta flag, thanks.

lwrubel:
I took the sleep statement out, since I have also gotten errors occasionally with it in. Just ran a couple of tests on stage and it was fine (after having an error earlier this afternoon.)

The sdr-ingest and bagit-errors aren't related to the cocinization--they happen on stage and QA. Is there something the test could be watching for in the UI? It's not clear to me what that might be.

spec/features/create_virtual_object_spec.rb
jcoyne: Yay, we caught a bug! 🐛 🐞 🥾

jcoyne (approved):

PR #354, opened 2022-02-25
[HOLD?] Preassembly with every access

jcoyne:
## Why was this change made?

This tests preassembly accessioning with every access.  This was able to detect https://github.com/sul-dlss/common-accessioning/issues/843 in a way that no single suite can with the tools we have today (see https://github.com/sul-dlss/cocina-models/issues/313)

Unfortunately, this adds about 20min to the run time and as a bonus you get 13 more emails. 🎉

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -105,6 +105,8 @@ Metrics/BlockLength:
     - it
     - context
     - scenario
+  Exclude:
+    - 'spec/features/create_preassembly_image_spec.rb'
 
 Metrics/MethodLength:
   Max: 20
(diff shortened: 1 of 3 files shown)


jcoyne:
@justinlittman I wonder if the changes to cocina-models in https://github.com/sul-dlss/cocina-models/pull/314 means we could have caught the bug in https://github.com/sul-dlss/common-accessioning/issues/843 without having to merge this (and adding a long time to the test suite)?

justinlittman:
Yes, with that test would have caught the bug. However, whether that obviates the need for this PR is a separate question.

mjgiarlo:
@jcoyne once https://github.com/sul-dlss/infrastructure-integration-test/pull/357 is merged and you rebase on `main`, the new CI checks should be good to go here.

ndushay:
Is this to be HOLD until we discuss?  If so, could that be added to the PR title?  Thanks.

PR #358, opened 2022-02-26
Test tweaks.

justinlittman:
Prefer if they pass.

### DIFF
--- a/spec/features/create_apo_spec.rb
+++ b/spec/features/create_apo_spec.rb
@@ -29,8 +29,6 @@ RSpec.describe 'Use Argo to create an APO and verify new objects inherit its rig
     # make sure APO is registered
     apo_druid = find_table_cell_following(header_text: 'DRUID').text
     expect(page).to have_content "APO #{apo_druid} created."
-    object_type_element = find_table_cell_following(header_text: 'Object type')
-    expect(object_type_element.text).to eq('adminPolicy')
 
     # wait for accessioningWF to finish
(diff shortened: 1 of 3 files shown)


spec/features/create_apo_spec.rb
```diff
@@ -29,8 +29,6 @@
     # make sure APO is registered
     apo_druid = find_table_cell_following(header_text: 'DRUID').text
     expect(page).to have_content "APO #{apo_druid} created."
-    object_type_element = find_table_cell_following(header_text: 'Object type')
```
justinlittman, line 32: This has been removed from table.

spec/features/create_object_h2_spec.rb
```diff
@@ -102,7 +102,7 @@
       click_link
     end
     # check embargo date
-    embargo_date = DateTime.now.to_date >> 6
+    embargo_date = DateTime.now.getutc.to_date >> 6
```
justinlittman, line 105: Otherwise you end up with the wrong date certain times of the day.

spec/features/create_virtual_object_spec.rb
```diff
@@ -14,6 +14,9 @@
   scenario do
     ensure_token
 
+    # Create virtual object
```
justinlittman, line 17: Avoid a race condition with ingest into preservation. Maybe.

jcoyne (approved):

PR #367, opened 2022-03-07
changing embargo now requires versioning

peetucket:

This recent change (sul-dlss/argo#3132) requires updating the tests to open a version before changing embargo

Note: the re-index seems to be flaky for me (i.e. it sometimes does not show the reindex message making the test get hung up here: https://github.com/sul-dlss/infrastructure-integration-test/blob/main/spec/support/page_helpers.rb#L24 which is called from here: https://github.com/sul-dlss/infrastructure-integration-test/blob/main/spec/features/create_object_h2_spec.rb#L120-L121)

### DIFF
--- a/spec/features/create_object_h2_spec.rb
+++ b/spec/features/create_object_h2_spec.rb
@@ -112,6 +112,13 @@ RSpec.describe 'Use H2 to create an object', type: :feature do
     # change embargo date
     new_embargo_date = Date.today + 3
     visit "#{Settings.argo_url}/view/#{bare_druid}"
+    # open a new version so we can manage embargo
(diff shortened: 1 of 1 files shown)


ndushay (approved):
does the version get closed?

peetucket:
> does the version get closed?

Yes, it gets closed by the H2 UI changes just below -- if you close it right after the embargo update, it will intefere with the H2 UI changes

I did note some flakiness which I wanted others to check out (basically test gets hung up on a reindex not showing the message sometimes, unless i click manually in my test browser)

peetucket:
Justin L said it passes for him - merging

PR #368, opened 2022-03-07
Fix faraday config for ETD spec.

justinlittman:
Faraday auth config changed in v2.


justinlittman:
Passes, so merging.

PR #371, opened 2022-03-09
Revert "Update dependencies for cocina-models update"

jmartin-sul:
Reverts sul-dlss/infrastructure-integration-test#370

d'oh, i realized as i merged that this the passing build was not a good measure of this PR working, will re-create the update PR with fixes after this undoes my bad merge


jmartin-sul:
@justinlittman says no need to revert, "will test it as part of testing the cocina update."

PR #379, opened 2022-03-18
Switch source of random text

### LINKED ISSUE #361: Implement alternative approach for creating unique titles
Currently using random-word gem to generate titles, which can accidentally generate offensive title strings. Solution needs to create titles that are unique so that they can be searched for in the course of the test.

lwrubel:
Resolves #361 to generate random text using a different source of words than the RandomWord gem.


lwrubel:
See issue #361 for example screenshots. If there are more you'd like to see, let me know.

jcoyne (approved):

mjgiarlo:
@lwrubel Thanks for working on this and coming up with such a thoughtful solution!

PR #388, opened 2022-03-28
Specifying bulk actions text to click

lwrubel:
Capybara needed more specific guidance on which of the renamed "Bulk Actions" links to click on or look for. Successfully ran on stage.

### DIFF
--- a/spec/features/edit_bulk_tags_spec.rb
+++ b/spec/features/edit_bulk_tags_spec.rb
@@ -22,8 +22,10 @@ RSpec.describe 'Use Argo to edit administrative tags in bulk', type: :feature do
     # Grab top three druids for testing bulk tag operation
     bulk_druids = all('dd.blacklight-id').take(number_of_druids).map(&:text)
 
-    click_link 'Bulk Actions'
-    expect(page).to have_content 'Bulk Actions'
+    within('.search-widgets') do
+      click_link 'Bulk Actions'
+    end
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -200,8 +196,7 @@ GEM
     zeitwerk (2.5.4)
 
 PLATFORMS
-  ruby
-  x86_64-darwin-19
+  x86_64-darwin-20
```
ndushay, line 199: I think you have to do the magic to not update platforms - did you do "bundle update" instead of "bundle install" ?

spec/features/edit_bulk_tags_spec.rb
```diff
@@ -22,8 +22,10 @@
     # Grab top three druids for testing bulk tag operation
     bulk_druids = all('dd.blacklight-id').take(number_of_druids).map(&:text)
 
-    click_link 'Bulk Actions'
-    expect(page).to have_content 'Bulk Actions'
+    within('.search-widgets') do
+      click_link 'Bulk Actions'
+    end
+    expect(page).to have_selector 'h1', text: 'Bulk Actions'
```
ndushay, line 28: This looks fine.

ndushay (changes requested):

ndushay (approved):

Gemfile.lock
lwrubel: @ndushay thanks, have removed this file from the PR

PR #391, opened 2022-04-05
update tests to handle Argo's lazily loaded sections

jmartin-sul:

A couple of tests broke due to the newly added lazy loading of the Events and Technical Metadata sections (Argo view page).

In both cases, I discovered that even if the section is expanded, the content doesn't actually load till it scrolls into view.  Since the Capybara-controlled browser only scrolls far enough to get a clicked element into view, clicking these section headers doesn't actually scroll the content into view (the section header starts out below the end of the viewport, so clicking it puts only the section header into the viewport).

Sort of lucky that I figured it out -- was tipped off when I put my cursor into the Capybara-controlled browser's viewport to scroll down a bit and see what Capybara saw at the moment in the events section (since all the events seemed to be there when i checked after the test failed)... and the moment I did that the previously failing test found what it was looking for and passed.

This fix feels a little hacky, so I'm happy to handle this differently if there's a better way.

N/A

### DIFF
--- a/spec/features/sdr_deposit_spec.rb
+++ b/spec/features/sdr_deposit_spec.rb
@@ -34,6 +34,12 @@ RSpec.describe 'SDR deposit', type: :feature do
     # Tests existence of technical metadata
     expect(page).to have_content 'Technical metadata'
     click_button 'Technical metadata'
+
+    # this is a hack that forces the techMD section to scroll into view; the section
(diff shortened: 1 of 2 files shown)


jmartin-sul:
see also https://github.com/sul-dlss/argo/pull/3366 and https://github.com/sul-dlss/argo/pull/3361

jcoyne (approved):

PR #417, opened 2022-04-25
wait for PURL page to be ready

peetucket:

The preassembly image test failed a couple times because the PURL page didn't seem to be quite ready when it tried to visit it.  Refreshing the test browser allowed the test to continue and pass.  This will give us a second to wait for that PULR page and should make this test less flaky.  I am not sure why it's not ready when the status in Argo says it is accessioned (it should be).

### DIFF
--- a/spec/features/create_preassembly_image_spec.rb
+++ b/spec/features/create_preassembly_image_spec.rb
@@ -139,6 +139,7 @@ RSpec.describe 'Create and re-accession object via Pre-assembly', type: :feature
 
     # This section confirms the object has been published to PURL and has a
     # valid IIIF manifest
+    sleep 1 # noticed the PURL page took a bit longer to be ready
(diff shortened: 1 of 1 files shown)


jcoyne:
@peetucket has anyone else been able to duplicate this error? I've never seen this one before.

jcoyne:
I've asked what may be causing this delay on the access slack channel: https://stanfordlib.slack.com/archives/C060ZTZRV/p1650916909627579

peetucket:
Ok.  I just hit it twice in stage today.  Haven't seen it previously.

jcoyne:
@peetucket yeah, me too. Never seen it before, but hitting it several times today.  I'm not sure that a 1 second wait is enough time even.

jcoyne (approved):

PR #443, opened 2022-06-21
Press Enter key to submit search term

jmartin-sul:

before this fix, `create_object_h2_spec.rb` consistently failed with `Capybara::Ambiguous:`  `Ambiguous match, found 2 elements matching visible button "Search" that is not disabled` on the parts of this spec that i modified.

after this fix, `create_object_h2_spec.rb` consistently passes.

There are multiple buttons on the page that contain the text "Search" (the other is the "Released to Searchworks" facet).  But the current locator for finding and filling in the search box still finds exactly one element, and pressing the Enter key in it after filling it in submits the search form successfully.

(Finding based on exact text would also be tricky: I noticed when debugging that the actual text of the search box submission button, from the perspective of Capybara::Node::Element#text is "Search search icon" because of alt text and icon, even when the text is limited to what Capybara considers visible.)

n/a

### DIFF
--- a/spec/features/create_object_h2_spec.rb
+++ b/spec/features/create_object_h2_spec.rb
@@ -87,8 +87,7 @@ RSpec.describe 'Use H2 to create an object', type: :feature do
 
     # Opens Argo and searches on title
     visit Settings.argo_url
-    fill_in 'Search...', with: item_title
-    click_button 'Search'
+    find_field('Search...').send_keys(item_title, :enter)
     reload_page_until_timeout!(text: 'v1 Accessioned')
 
     # check Argo facet field with 6 month embargo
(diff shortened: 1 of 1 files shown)


spec/features/create_object_h2_spec.rb
```diff
@@ -87,8 +87,7 @@
 
     # Opens Argo and searches on title
     visit Settings.argo_url
-    fill_in 'Search...', with: item_title
-    click_button 'Search'
+    find_field('Search...').send_keys(item_title, :enter)
```
jmartin-sul, line 90: [`fill_in`](https://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Actions:fill_in) returns the element that was filled in, so if people think it reads better, i could instead do something like:

```ruby
search_entry_field = fill_in 'Search...', with: item_title
search_entry_field.send_keys(:enter)
```

the first param to `send_keys` here just inputs the given string as text.  the second param causes an `Enter` key press.  ([`send_keys`](https://www.rubydoc.info/github/jnicklas/capybara/Capybara%2FNode%2FElement:send_keys) is pretty nice, it takes an arbitrary list of params, representing a sequence of keystrokes, and a simultaneous key combo can be given by an array param, e.g. `[:control, 'a']` to select all text in a textbox (depending on the OS, for that last example).

mjgiarlo (approved):

PR #444, opened 2022-06-21
label change for pre-assembly

peetucket:

Because of https://github.com/sul-dlss/pre-assembly/pull/1012, we need to update our integration tests for pre-assembly.

Wait until that PR is merged


justinlittman (approved):

ndushay:
@peetucket #1012 is merged ... not clear if this should be merged now or friday afternoon so it's in place for dep updates on Monday morning.

(at the moment, deployed environments may have particular preassembly branches for Andrew's testing)

peetucket:
unless folks are actively running integration tests against older branches of pre-assembly that are deployed, i'd suggest we just merge this now to avoid further confusion

note that since sul-dlss/pre-assembly#1012 is now merged and it has a database migration, if you deploy main and then try and re-deploy a dev branch started before #1012 was merged, it will blow up (unless you rebase main onto the dev branch)

PR #450, opened 2022-07-01
Updates specs for changes to druid after registering.

justinlittman:
After registering, bare druids are displayed instead of prefixed druids.


ndushay (approved):
this is fine;  i think we just use `bare_druid` in general, if you want to consider changing it to that.

PR #457, opened 2022-07-07
Update access_web_archive_spec to verify url in purl xml

### LINKED ISSUE #456: finish was seed accession test
Spoke to Andrew;  he would like the was seed accession test (part of `accession_web_archive_spec`) to include checking the purl page xml for the MODS location/url[displayLabel="Archived website"] that matches the full URL.  The spec has `url_in_wayback` variable, but the test should look for the wayback prefix, etc.

image

So here the `url_in_wayback` is "https://library.stanford.edu/department/digital-library-systems-and-services-dlss/about-us" but the URL we are looking for is "#{Settings.something}/*/#{url_in_wayback}"

You can get to the purl xml like this (or something similar) from the create_preassembly_image_spec;  create_object_h2_spec checks purl xml as well.

```
    visit "#{Settings.purl_url}/#{druid.delete_prefix('druid:')}"
    # wait for the PURL name to be published by checking for collection name
    reload_page_until_timeout!(text: collection_name)
```

aaron-collier:

Closes #456

### DIFF
--- a/config/settings/staging.yml
+++ b/config/settings/staging.yml
@@ -14,4 +14,3 @@ was_registrar:
   jobs_directory: '/was_unaccessioned_data/jobs'
   url: 'https://was-registrar-app-stage.stanford.edu'
   username: 'was'
-swap_url: 'https://swap-stage.stanford.edu'
--- a/config/settings/staging.yml
+++ b/config/settings/staging.yml
@@ -14,3 +14,4 @@ was_registrar:
   jobs_directory: '/was_unaccessioned_data/jobs'
   url: 'https://was-registrar-app-stage.stanford.edu'
   username: 'was'
+swap_url: 'https://swap-stage.stanford.edu'
--- a/spec/features/accession_web_archive_spec.rb
+++ b/spec/features/accession_web_archive_spec.rb
@@ -23,6 +23,7 @@ RSpec.describe 'Use was-registrar-app, Argo, and pywb to ensure web archive craw
     end
   end
   let(:url_in_wayback) { 'https://library.stanford.edu/department/digital-library-systems-and-services-dlss/about-us' }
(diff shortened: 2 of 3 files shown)


spec/features/accession_web_archive_spec.rb
```diff
@@ -88,5 +89,12 @@
     expect(page).to have_link('thumbnail.jp2')
     expect(page).to have_content('image/jp2')
     expect(page).to have_content('400 px')
+
+    # Verify that the purl XML includes the proper archived website URL
+    purl_xml = "#{Settings.purl_url}/#{seed_druid}.xml"
```
ndushay, line 94: there's a way we get the purl_xml from dor-services-app so it's faster ... is this doing that?  See https://github.com/sul-dlss/infrastructure-integration-test/blob/main/spec/support/public_xml_helpers.rb 

I'm surprised this doesn't time out, tbh.  maybe because it's the purl xml, and not the purl??

ndushay (approved):
approving but with a question

spec/features/accession_web_archive_spec.rb
aaron-collier: Hmm.. I'm happy to make that change.

config/settings/staging.yml
```diff
@@ -14,3 +14,4 @@ was_registrar:
   jobs_directory: '/was_unaccessioned_data/jobs'
   url: 'https://was-registrar-app-stage.stanford.edu'
   username: 'was'
+swap_url: 'https://swap-stage.stanford.edu'
```
ndushay, line 17: won't we need this for qa as well??

ndushay (approved):
one more tweak needed

mjgiarlo (approved):

PR #464, opened 2022-07-21
new goobi integration test

### LINKED ISSUE #459: objects from goobi need integration test
We would have to start with data for an object _AS IF_ we got it from goobi, with the appropriate workflow step being kicked off.   

Since goobi is another special snowflake that we can break inadvertently, it behooves us to have an integration test to show that given the handoff from goobi to accessioning, everything runs properly and the cocina structural is as expected and object accessions cleanly and looks right in Argo.

To do:
- [ ] figure out why the "integration" goobi user is not seeing the items registered in argo even when the correct project is added
- [ ] figure out how to upload files in goobi and click the various controls that need to be clicked to advance the object there
- [ ] match goobi versions in qa and prod (intranda task)

peetucket:

Fixes #459 - adds new integration that runs an image to be accessioned through goobi-stage.

Todo:
- [x] Needs to have README info updated about how to get goobi-stage login info (see https://github.com/sul-dlss/shared_configs/pull/1831)
- [x] Need to verify it all works correctly once stacks-stage is working again

yes


spec/features/create_goobi_image_spec.rb
```diff
@@ -0,0 +1,115 @@
+# frozen_string_literal: true
+
+require 'druid-tools'
```
mjgiarlo, line 3: Is this dependency still used below or can we nix it?

spec/features/create_goobi_image_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    # wait for object to be registered
+    expect(page).to have_text 'Items successfully registered.'
+
+    bare_object_druid = find('table a').text
+    druid = "druid:#{bare_object_druid}"
+
+    # puts druid # useful for debugging
+
+    # wait to be sure goobiWF has finished running and goobi has time to process the incoming object
+    sleep 2
```
mjgiarlo, line 53: I'm curious which line below fails if we don't sleep here, and wondering if we can use a reload loop instead. It might also be a six-of-one-vs-half-dozen-of-other situation.

spec/features/create_goobi_image_spec.rb
peetucket: good catch, it was copypasta and I removed it

spec/features/create_goobi_image_spec.rb
peetucket: basically if goobiWF has not finished running yet, then the goobi system will not yet aware of the object, and it won't be found (because the main task of goobiWF is to make a POST call out to the goobi system to register the object there)

instead of waiting, we could visit the detail page in argo and ensure the workflow is completed before proceeding, but unless it sometimes takes longer and this becomes flaky, that feels like more extra work (and likely just if not more actual time passed) than just pausing for a bit

mjgiarlo (approved):

PR #465, opened 2022-07-21
Stub out expected config values

mjgiarlo:

Do this so developers have a sense of which things they might expect to override in local config not tracked under version control.

No

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -36,3 +36,8 @@ dor_services:
 sunet:
   id: ~ # NOTE: *without* @stanford.edu!
   password: ~
+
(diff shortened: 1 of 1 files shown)


config/settings.yml
```diff
@@ -36,3 +36,8 @@ dor_services:
 sunet:
   id: ~ # NOTE: *without* @stanford.edu!
   password: ~
+
+# the URL and username/password used to login to goobi (stored in vault)
```
justinlittman, line 40: Turns out they were not put into vault. See https://stanfordlib.slack.com/archives/C0RK5EM9N/p1658427869896549

justinlittman (changes requested):

PR #466, opened 2022-07-22
fix issues with goobi integration test

peetucket:

When uploading an image via the goobi UI during the test, we need to wait a bit for the upload to finish before proceeding with the test.

### DIFF
--- a/spec/features/create_goobi_image_spec.rb
+++ b/spec/features/create_goobi_image_spec.rb
@@ -60,7 +60,7 @@ RSpec.describe 'Create and accession object via Goobi', type: :feature,
     click_button 'Log in'
 
     # find the new object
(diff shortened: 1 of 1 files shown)


spec/features/create_goobi_image_spec.rb
```diff
@@ -60,14 +60,18 @@
     click_button 'Log in'
 
     # find the new object
-    expect(page).to have_selector('h2', text: 'Home page')
+    expect(page).to have_text('Home page')
     click_link 'My tasks'
```
peetucket, line 64: noticed that it is now always in an `<h2>` tag (maybe depends on the version of goobi installed)...relax this expectation a bit

jcoyne (approved):

PR #468, opened 2022-07-29
GIS integration test

### LINKED ISSUE #16: gis-robots accessioned objects need integration tests
I'm not sure how a GIS object is created and what makes it go thru GIS robots.  Kim Durante or @mejackreed might be able to help.

- creating a new object in SDR (that runs through gis-robots)
- updating a GIS object in SDR

The test should mimic what is done, and should check for GIS object-ness in Argo. All the way through accessioning, if that is possible.

peetucket:

Add a new integration test for accessioning GIS objects.  Fixes #16 

Note: there is no geoserver-qa, so the test fails part way if run in the QA environment (e.g. see https://argo-qa.stanford.edu/view/druid:sb443ft9972).  

Todo:
- [ ] update this object too?

### DIFF
--- a/spec/features/create_gis_object_spec.rb
+++ b/spec/features/create_gis_object_spec.rb
@@ -1,0 +1,105 @@
+# frozen_string_literal: true
+
+# NOTE: this spec will be skipped unless run on staging, since there is no geoserver-qa
+RSpec.describe 'Create and accession GIS object', type: :feature,
+                                                  if: ENV.fetch('SDR_ENV', 'staging') == 'staging' do
+  let(:start_url) { "#{Settings.argo_url}/registration" }
+  let(:project_name) { 'GIS Integration Test' }
+  let(:source_id_random_word) { "#{random_noun}-#{random_alpha}" }
+  let(:source_id) { "gis-integration-test:#{source_id_random_word}" }
+  let(:label_random_words) { random_phrase }
+  let(:object_label) { "gis integration test #{label_random_words}" }
+  let(:collection_name) { 'Integration Test Collection - GIS' }
+  # this APO must exist on argo-stage and argo-qa and have "Integration Test Collection - GIS" collection available
+  let(:apo_name) { 'APO for GIS' }
+
+  before do
+    authenticate!(start_url: start_url,
+                  expected_text: 'Register DOR Items')
+  end
+
+  scenario do
+    # register new GIS object
+    select apo_name, from: 'Admin Policy'
+    select collection_name, from: 'Collection'
+    select 'file', from: 'Content Type'
+    fill_in 'Project Name', with: project_name
+    fill_in 'Source ID', with: source_id
+    fill_in 'Label', with: object_label
+
+    click_button 'Register'
+
+    # wait for object to be registered
+    expect(page).to have_text 'Items successfully registered.'
+
+    bare_object_druid = find('table a').text
+    druid = "druid:#{bare_object_druid}"
+
+    # # puts druid # useful for debugging
+
+    # Go to kurma server and copy test content to the druid folder so it can be accessioned
+    test_data_source_folder = File.join(Settings.gis.robots_content_root, 'integration_test_data')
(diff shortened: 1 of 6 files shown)


spec/support/page_helpers.rb
```diff
@@ -41,4 +45,5 @@ def reload_page_until_timeout!(text:, as_link: false, table: nil, with_reindex:
   # rubocop:enable Metrics/CyclomaticComplexity
   # rubocop:enable Metrics/MethodLength
   # rubocop:enable Metrics/PerceivedComplexity
+  # rubocop:enable Metrics/ParameterLists
 end
```
peetucket, line 49: allow us to use this method with a specific element on the page (useful for verifying specific workflow completion)

spec/features/create_gis_object_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+    expect(page).to have_text 'Items successfully registered.'
+
+    bare_object_druid = find('table a').text
+    druid = "druid:#{bare_object_druid}"
+
+    # # puts druid # useful for debugging
+
+    # Go to kurma server and copy test content to the druid folder so it can be accessioned
+    test_data_source_folder = File.join(Settings.gis.robots_content_root, 'integration_test_data')
+    test_data_destination_folder = File.join(Settings.gis.robots_content_root, bare_object_druid, 'temp')
```
peetucket, line 40: note: this integration test data is currently just sitting on the server...another alternative is to place it in this codebase and then SCP up to the server each time it is run

spec/support/page_helpers.rb
```diff
@@ -21,6 +23,8 @@ def reload_page_until_timeout!(text:, as_link: false, table: nil, with_reindex:
           break if page.has_link?(text, wait: wait_time)
         elsif table
           break if page.find(:table_row, table).text.match?(text)
+        elsif selector
+          break if page.has_selector?(selector, text: text, wait: wait_time)
         else
```
peetucket, line 28: this adds some extra functionality to this helper method to allow us to wait for text in a specific selector, useful for waiting for a specific workflow to complete

ndushay (commented):
I'm about to put in a PR to rename the "create_xxx" spec files.   What do you think about calling the file `gis_accessioning_spec.rb` ?

jcoyne:
@ndushay I love the rename idea. If there are no other changes for this PR, you could merge this PR first and then do that rename in your PR.

ndushay (changes requested):
Trying to run this on stage, there's an error in the normalize-data step.  See https://argo-stage.stanford.edu/view/druid:tj263wq9366

```
  1) Create and accession GIS object is expected not to have visible css ".alert-danger"
     Failure/Error: expect(page).not_to have_css('.alert-danger', wait: 0)
       expected not to find visible css ".alert-danger", found 1 match: "Error: normalize-data : 502 Bad Gateway"
     # ./spec/support/page_helpers.rb:34:in `block (2 levels) in reload_page_until_timeout!'
     # ./spec/support/page_helpers.rb:11:in `loop'
     # ./spec/support/page_helpers.rb:11:in `block in reload_page_until_timeout!'
     # ./spec/support/page_helpers.rb:10:in `reload_page_until_timeout!'
     # ./spec/features/create_gis_object_spec.rb:57:in `block (2 levels) in <top (required)>'
```

image

https://app.honeybadger.io/projects/52899/faults/87111228

Is this working in production?   Do we need https://spatialreference.org/ to allow connections from our gis-robots stage server?

peetucket:
1. I renamed the spec file.
2. I noticed the project tag was a bit different than other intergration test project tags, so I adjusted it slightly (not important for the test, but useful for finding the objects later)
3. The `https://spatialreference.org/` error is new as of this week and I don't think is under our control.  That appears to be an external domain/website that is actually down right now, which will block production accessioning of GIS objects too.

ndushay (approved):
Now that the external dependency is available, I got this to run.  Yay!

PR #473, opened 2022-08-09
output druids for "when tests go bad"

ndushay:

So we don't have to fuss with cut and paste from the browser window for tests in progress.

Also updated a few spec descriptions


justinlittman (approved):

jcoyne:
Thanks! This will be a great help!

PR #478, opened 2022-08-22
Adjusting h2 form input for sunetID lookup

lwrubel:
New SunetID lookup field needs a moment to display and then correct Add button needs to be clicked. Now runs on stage and QA.

### DIFF
--- a/spec/features/h2_object_creation_spec.rb
+++ b/spec/features/h2_object_creation_spec.rb
@@ -32,6 +32,9 @@ RSpec.describe 'Use H2 to create an object', type: :feature do
     # Adds user to depositor field
     click_button '+ Add another depositor'
     fill_in 'lookup SunetID', with: AuthenticationHelpers.username
+    within '.participant-overlay' do
+      click_button 'Add'
+    end
 
     click_button 'Deposit'
     expect(page).to have_content(collection_title)


spec/features/h2_object_creation_spec.rb
```diff
@@ -31,7 +31,12 @@
 
     # Adds user to depositor field
     click_button '+ Add another depositor'
+    # wait a second for the SunetID search field to display
+    sleep 1
```
jcoyne, line 35: I don't understand why this is necessary. Nothing around this changed.

jcoyne (changes requested):

spec/features/h2_object_creation_spec.rb
jcoyne: I pulled this branch and commented this line out and the test still works, so can it be removed?

lwrubel: Thanks for testing it and confirming it worked without the sleep line. Have removed. Also at this step, do you know why I might sometimes getting the following and other times it successfully does a lookup? 
Screen Shot 2022-08-22 at 11 12 13 AM

jcoyne: I saw that too once. This is also a problem that we've seen in production before too. I'm not sure they are connected.

jcoyne (approved):

PR #481, opened 2022-08-25
change druids searched for in tag editing test

peetucket:

The existing integration tests search for the first three druids in all of Argo, one of them currently has a date validation problem.  This just uses a specific collection to pull the top three druids from - hopefully will have cleaner data (and it works right now).


jcoyne:
There is no longer any validation problem. Should we close this?

peetucket:
seems reasonable to me

PR #507, opened 2022-10-28
Preassembly hfs test

peetucket:

New integration test for hierarchical file layout

~~HOLD for sul-dlss/pre-assembly#1104, which needs to be merged for this test to work~~

### DIFF
--- a/spec/features/preassembly_image_accessioning_spec.rb
+++ b/spec/features/preassembly_image_accessioning_spec.rb
@@ -84,6 +84,8 @@ RSpec.describe 'Create and re-accession image object via Pre-assembly' do
     wait_for_download
     yaml = YAML.load_file(download)
     expect(yaml[:status]).to eq 'success'
+    # delete the downloaded YAML file, so we don't pick it up by mistake during the re-accession
+    delete_download(download)
 
     # ensure Image files are all there, per pre-assembly, organized into specified resources
     visit "#{Settings.argo_url}/view/#{druid}"
--- a/config/settings/qa.yml
+++ b/config/settings/qa.yml
@@ -8,6 +8,7 @@ gis:
 # for purl_url: there is no purl for qa;  sul-h2-qa uses stage ...
 purl_url: 'https://sul-purl-stage.stanford.edu'
 preassembly_bundle_directory: '/dor/staging/integration-tests/image-test'
+preassembly_hfs_bundle_directory: '/dor/staging/integration-tests/hierarchical-file-test'
 preassembly_host: 'sul-preassembly-qa.stanford.edu'
 preassembly_url: 'https://sul-preassembly-qa.stanford.edu'
 sdrapi_url: 'https://sdr-api-qa.stanford.edu'
--- a/config/settings/stage.yml
+++ b/config/settings/stage.yml
@@ -7,6 +7,7 @@ gis:
 h2_url: 'https://sul-h2-stage.stanford.edu'
 purl_url: 'https://sul-purl-stage.stanford.edu' # from shared_configs for sul-h2-stage
 preassembly_bundle_directory: '/dor/staging/integration-tests/image-test'
+preassembly_hfs_bundle_directory: '/dor/staging/integration-tests/hierarchical-file-test'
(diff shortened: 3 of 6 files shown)


spec/features/preassembly_image_accessioning_spec.rb
```diff
@@ -82,6 +82,8 @@
     wait_for_download
     yaml = YAML.load_file(download)
     expect(yaml[:status]).to eq 'success'
+    # delete the downloaded YAML file, so we don't pick it up by mistake during the re-accession
+    delete_download(download)
```
peetucket, line 87: i noticed the way we look for downloaded YAML files was just picking the first, yet in a re-accession we will have two until we cleanup ... so safer to cleanup the first after it's done before we start the re-accession to avoid false positives

jmartin-sul (changes requested):
this could use a rebase, but for quick testing purposes, i rebased this branch on my branch from #521 (because that made authN less annoying, while not picking up the more substantial refactoring of #517 that's now on `main`, which would require changes to at least the `reload_page_until_timeout!` calls).

and while the new test in this PR passed after that rebase, and created a hierarchical object that got preserved, i noticed an issue when digging into things.

i was able to download the zip file, and it included all the file hierarchy i'd expect, including `images/subdir/image.jpg`.  maybe this zip is made from stacks content?  i'm not sure about that without looking.

unfortunately, that nested sibdir doesn't seem to be preserved in the moab.  when i recursively listed the moab, i did see `v0001/data/content/images`.  however, i saw no subdirs under that, and i get a 404 when i try to download `images/subdir/image.jpg` by itself from preservation via argo.

here's the test object that the run created: https://argo-stage.stanford.edu/view/sy413pc2719

its moab is here: `/services-disk-stage/store2/sdr2objects/sy/413/pc/2719/sy413pc2719/`

so, i think there are two things to do here, one in the scope of this PR, one not
* add some sort of check to this new test to make sure that the moab is correct and complete.
* preserve the full hierarchy in the generated moab (probably requires tweaking moab-versioning gem code?)

i didn't really dig into the code here yet (though i did notice the replication check, and i liked that -- a nice follow-on refactoring might be to extract that now common code to a shared method, since `preassembly_image_accessioning_spec.rb` does the same).

andrewjbtw:
@jmartin-sul I think what you're seeing in Argo is a URL encoding problem.

I get this link from Argo:
`https://argo-stage.stanford.edu/items/druid:sy413pc2719/files/config%2Fsettings.yml/preserved?version=2`

Switching a slash for the `%2F` gets a working link:
`https://argo-stage.stanford.edu/items/druid:sy413pc2719/files/config/settings.yml/preserved?version=2`

I thought the Argo links had been patched to handle the slashes but maybe more work is needed there.

jmartin-sul:
@andrewjbtw nice find, thanks!  I guess I didn't give everything with that level of nesting a look, so that's reassuring that at least some of it is handled correctly.  I'd missed that `content/config/settings` and its contents were correctly preserved in the Moab when I was looking at this on Friday evening.  But I just looked at the Moab on its storage root again, and that subdir and its contents are indeed present (which makes sense, since the fixed download link you provided works).

But I'm still confused as to why `images/subdir/image.jpg` doesn't appear in the moab...

Ah, ok, now I think i get it.  It appears that `images/subdir/image.jpg` and `images/image.jpg` have the same file content.  So this is probably just moab-versioning (correctly) de-duping the file content based on file checksums.  If I modify the `images/subdir/image.jpg` preservation download URL to use slashes as you did for the test files you mentioned, I'm able to download the image file that I couldn't get on Friday.

jmartin-sul:
Ok, I closed the moab-versioning ticket with a bit of explanation as to where I was off-base.

Might still be nice to augment this test with something that checks the download links.  Maybe specifically for the file that confused me, with a comment in the test indicating that by downloading the file and checking its checksum we're 1) testing the download link, 2) testing preservation of something deeply nested in the folder structure, 3) testing preservation of content that moab-versioning will try to de-dupe.  Wouldn't hold up merging this PR over that, though.  Will dismiss my "request for changes" and update my review after lunch and the 1 pm meeting.

mjgiarlo (approved):

PR #510, opened 2022-11-02
Configure for Cardinal Key

edsu:

These two settings are needed (for some) to not be barraged with confirmation dialogs about the certificate from Cardinal Key:

- `security.enterprise_roots.enabled` set to `true`
- `security.default_personal_cert` set to `Select Automatically`

See: https://uit.stanford.edu/service/cardinalkey/known-issues

No.


justinlittman (approved):
WIN!!!

PR #512, opened 2022-11-07
Update authentication to add brower trust button step

mjgiarlo:

This is a new step, apparently. Also, the nearby sleep appears no longer to be needed. Leaving commented out in case someone else runs into auth issues in the near future, as a clue for them to follow.

no

### DIFF
--- a/spec/support/authentication_helpers.rb
+++ b/spec/support/authentication_helpers.rb
@@ -27,7 +27,7 @@ module AuthenticationHelpers
     # We're at the Stanford login page
     fill_in 'SUNet ID', with: username
     fill_in 'Password', with: password
-    sleep 1
+    # sleep 1 # mjgiarlo found he didn't need this on 2022/11/07 so commenting out
     click_button 'Login'
 
     if Settings.automatic_authentication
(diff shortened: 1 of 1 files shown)


spec/support/authentication_helpers.rb
```diff
@@ -48,6 +48,8 @@ def submit_credentials(expected_text = '')
         click_button 'Send Me a Push'
       end
     end
+
+    click_button 'Yes, trust browser'
```
ndushay, line 52: i wonder if this will be needed for all of us, or if it will cause tests to fail for some of us.  The convoluted logc of L33-50 was for a similar situation.

ndushay (commented):
I'm not comfortable approving with the new change as I'm not sure we all need it.  I'm about to try it now.

spec/support/authentication_helpers.rb
mjgiarlo: Sounds good, thx for testing

ndushay (approved):
My experience:

I tested using Firefox.

- having this change included (using `authentication-update` branch or using `main` branch behaved the same.
- this change did not fooch the tests.
- I still had to manually click "Yes, trust browser" button 
- I only had to click it once when running multiple tests.

So ... it's okay with me if it actually helps someone?

I wonder if the previous lines are still useful

I also wonder why using firefox, it didn't find the button:

image

mjgiarlo:
@ndushay I too am using Firefox. There is a slight delay (~2-3s?) before the button is clicked, I've found. Thanks for the approval. I somehow think we're going to iterate on this every week for a while...

PR #514, opened 2022-11-11
[HOLD] Fix multi-test authentication bug and update webdrivers + rubocop

mjgiarlo:

More maintenance for our oft-neglected integration test suite:

* Fix authentication bug when running more than one test
  * This commit restores past logic for handling the situation when running more than one test, making sure not to reauthenticate when not needed.
* Unpin webdrivers gem
  * As far as I can tell, this is no longer necessary.
* Update rubocop config and switch to Ruby 3.1
  * Also, consistently use `has_text?` (vs. `has_content?`)---they are the same method. It does not matter which one we use, but it is better not to suggest they are doing something different.

Yes.


.rubocop.yml
```diff
@@ -5,7 +5,7 @@ require:
 inherit_from: .rubocop_todo.yml
 
 AllCops:
-  TargetRubyVersion: 3.0
+  TargetRubyVersion: 3.1
```
mjgiarlo, line 8: A wee bump as we march towards Ruby 3.1. Nice to reduce all the duplication in pre-3.1 hash syntax.

.rubocop.yml
```diff
@@ -378,3 +378,21 @@ RSpec/NoExpectationExample: # new in 2.13
...
+RSpec/SortMetadata: # new in 2.14
+  Enabled: true
+RSpec/Capybara/NegationMatcher: # new in 2.14
+  Enabled: true
+RSpec/Capybara/SpecificActions: # new in 2.14
+  Enabled: true
+RSpec/FactoryBot/ConsistentParenthesesStyle: # new in 2.14
+  Enabled: true
+RSpec/Rails/InferredSpecType: # new in 2.14
+  Enabled: true
```
mjgiarlo, line 398: Routine maintenance keeping up with Rubocop cops.

.rubocop_todo.yml
```diff
@@ -1,12 +1,7 @@
 # This configuration was generated by
 # `rubocop --auto-gen-config`
-# on 2022-08-02 17:23:52 UTC using RuboCop version 1.32.0.
+# on 2022-11-10 23:49:07 UTC using RuboCop version 1.38.0.
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
```
mjgiarlo, line 7: Check it out: an empty rubocop TODO file!

.rubocop_todo.yml
```diff
@@ -1,12 +1,7 @@
...
-
-# Offense count: 1
-Style/ClassVars:
-  Exclude:
-    - 'spec/support/authentication_helpers.rb'
```
mjgiarlo, line 12: Replaced the class variable below with a module attr.

Gemfile
```diff
@@ -15,10 +15,10 @@ gem 'faraday' # etds require POST request from registrar
...
-gem 'rubocop', '~> 1.24'
+gem 'rubocop'
 gem 'rubocop-rake'
-gem 'rubocop-rspec', '~> 2.1'
+gem 'rubocop-rspec'
 gem 'rubyXL' # for updating Excel spreadsheets
 gem 'sdr-client', '~> 0.97'
 gem 'selenium-webdriver'
-gem 'webdrivers', '~> 4.0'
+gem 'webdrivers'
```
mjgiarlo, line 24: This'll be fine.

spec/features/access_indexing_spec.rb
```diff
@@ -100,7 +100,7 @@ def choose_rights(view:, download: nil, location: nil, cdl: false)
   end
 
   # It takes a few milliseconds for the rights update to take
-  expect(page).to have_content(view_label(view: view, location: location, cdl: cdl))
+  expect(page).to have_text(view_label(view:, location:, cdl:))
```
mjgiarlo, line 103: Instead of inconsistent usage of has/ve_text and has/ve_content, pick one and be consistent.

spec/spec_helper.rb
```diff
@@ -59,6 +59,7 @@
 
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 RSpec.configure do |config|
+  config.include Capybara::DSL
```
mjgiarlo, line 62: This is required since we removed all the `type: :feature` jazz above.

spec/support/authentication_helpers.rb
```diff
@@ -1,25 +1,27 @@
 # frozen_string_literal: true
 
 module AuthenticationHelpers
-  mattr_accessor :username, :password
+  mattr_accessor :username, :password, :token
```
mjgiarlo, line 4: We already used mattrs in this module... why not use one more if it lets us ditch our only Rubocop TODO?

spec/support/authentication_helpers.rb
```diff
@@ -1,25 +1,27 @@
...
 
   def authenticate!(start_url:, expected_text:)
     # View the specified starting URL
     visit start_url
 
+    return if page.has_text?(expected_text, wait: Settings.post_authentication_text_timeout)
```
mjgiarlo, line 10: This is a check we used to make back in the old "automatic authentication" days (read: a couple hours ago) that only trips you up if 1. you're using Firefox which authenticates only once even when running multiple tests, and 2. running multiple tests.

mjgiarlo: This would be good for @justinlittman and @ndushay (who helped earlier) to test on their machines. I'd recommend:

```shell
bundle exec rspec spec/features/apo_creation_spec.rb spec/features/sdr_deposit_spec.rb
```

aaron-collier (approved):
LGTM - I won't merge until @ndushay and/or @justinlittman have a look since they were involved in one of the issues earlier. Note: This doesn't really change the behavior for me at all, the same usual tests fail in the suite, and then pass individually.

mjgiarlo:
> LGTM - I won't merge until @ndushay and/or @justinlittman have a look since they were involved in one of the issues earlier. Note: This doesn't really change the behavior for me at all, the same usual tests fail in the suite, and then pass individually.

If it's helpful to you, I can find time to pair on the failures you're seeing. ¯\_(ツ)_/¯ 🙂

justinlittman (approved):

spec/spec_helper.rb
ndushay: maybe indicate this in a comment?

ndushay (commented):
Mike:

The second test fails for me, whichever it is -- 

```
  1) Use Argo to create an APO and verify new objects inherit its rights 
     Failure/Error: click_button 'Yes, trust browser'
     
     Capybara::ElementNotFound:
       Unable to find button "Yes, trust browser" that is not disabled
     # ./spec/support/authentication_helpers.rb:28:in `submit_credentials'
     # ./spec/support/authentication_helpers.rb:10:in `authenticate!'
     # ./spec/features/apo_creation_spec.rb:15:in `block (2 levels) in <top (required)>'

```

mjgiarlo:
@ndushay OK, let's pair today if you have time.

spec/spec_helper.rb
mjgiarlo: @ndushay https://github.com/sul-dlss/infrastructure-integration-test/pull/515

PR #521, opened 2022-11-18
Authentication fixups

jmartin-sul:

* fix a typo
* more robust handling of different login workflow possibilities for different users
* accommodation of slow redirects through duo 2FA service when visiting item show page in argo

commit messages have a bit more detail about the motivation for the changes.

N/A, i think?

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -5,6 +5,8 @@ timeouts:
   events:
     poll_for: 240
     poll_interval: 2
+  etd:
+    embargo_poll_interval: 1
 
 browser:
   driver: firefox
--- a/spec/features/etd_creation_spec.rb
+++ b/spec/features/etd_creation_spec.rb
@@ -207,12 +207,13 @@ RSpec.describe 'Create a new ETD with embargo, and then update the embargo date'
     Timeout.timeout(Settings.timeouts.workflow) do
       loop do
         visit "#{Settings.argo_url}/view/#{prefixed_druid}"
-        break if page.has_text?("Embargoed until #{embargo_date.to_formatted_s(:long)}", wait: 1)
+        break if page.has_text?("Embargoed until #{embargo_date.to_formatted_s(:long)}",
+                                wait: Settings.timeouts.etd.embargo_poll_interval)
       end
     end
     expect(page).to have_text(dissertation_title)
     apo_element = find_table_cell_following(header_text: 'Admin policy')
-    expect(apo_element.first('a')[:href]).to have_text('druid:bx911tp9024') # this is hardcoded in hydra_etd app
+    expect(apo_element.first('a')[:href]).to include('druid:bx911tp9024') # this is hardcoded in hydra_etd app
     status_element = find_table_cell_following(header_text: 'Status')
     expect(status_element).to have_text('v1 Registered')
     click_link('etdSubmitWF')
--- a/spec/support/authentication_helpers.rb
+++ b/spec/support/authentication_helpers.rb
@@ -4,30 +4,20 @@ module AuthenticationHelpers
   mattr_accessor :username, :password, :token
 
   def authenticate!(start_url:, expected_text:)
+    ensure_username! # sunet is needed by some tests, even if the user doesn't have to enter user/pass for Stanford web authN
+
     # View the specified starting URL
     visit start_url
+    click_through_trust_browser_if_needed # for cardinal key users, straight to 2FA prompt, no login form
 
-    return if page.has_text?(expected_text, wait: Settings.post_authentication_text_timeout)
+    submit_credentials_if_needed
+    click_through_trust_browser_if_needed # for people who hit the login form, 2FA prompt comes after it's submitted
 
-    submit_credentials
-
-    using_wait_time(Settings.timeouts.capybara) do
-      # Once we see this we know the log in succeeded.
-      expect(page).to have_text(expected_text)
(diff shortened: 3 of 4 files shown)


jmartin-sul:
it'd be great to get people with the following setups to test this:
- [ ] a cardinal key user who stores their login creds in local config (i'm a cardinal key user, but i don't save my login info)
- [ ] a user who does not user cardinal key, and who does not store their login creds in local config (not sure we have any such users, so maybe this is moot?)
- [ ] a user who does not use cardinal key, who stores their login creds in local config (@mjgiarlo does this describe your setup?)

if this branch works for people whose workflows differ from mine in the above ways, i'll take it out of draft.

spec/features/etd_creation_spec.rb
```diff
@@ -207,12 +207,13 @@
     Timeout.timeout(Settings.timeouts.workflow) do
       loop do
         visit "#{Settings.argo_url}/view/#{prefixed_druid}"
-        break if page.has_text?("Embargoed until #{embargo_date.to_formatted_s(:long)}", wait: 1)
+        break if page.has_text?("Embargoed until #{embargo_date.to_formatted_s(:long)}",
+                                wait: Settings.timeouts.etd.embargo_poll_interval)
```
jmartin-sul, line 211: i made the new config value equal the old hardcoded value of 1 second.  for me, when this was 1 second, i noticed when watching the test browser that it just kept refreshing a duo URL, presumably a URL that stanford webauth bounced me to before letting me into argo, to see whether i needed another 2FA prompt.

when i upped this to 10 seconds for myself locally, the test no longer got stuck here when i ran it.

config/settings.yml
```diff
@@ -5,6 +5,8 @@ timeouts:
   events:
     poll_for: 240
     poll_interval: 2
+  etd:
+    embargo_poll_interval: 1
```
jmartin-sul, line 9: i was on the fence about using an existing timeout vs adding a new one.  this felt like the slightly safer bet, but happy to change.

spec/support/authentication_helpers.rb
```diff
@@ -4,30 +4,20 @@ module AuthenticationHelpers
   mattr_accessor :username, :password, :token
 
   def authenticate!(start_url:, expected_text:)
+    ensure_username! # sunet is needed by some tests, even if the user doesn't have to enter user/pass for Stanford web authN
```
jmartin-sul, line 7: after some recent changes, i was never prompted for my sunet, since i don't go through the username/pass login form anymore.  and since i don't store my login info locally, the ETD spec would get a blank username when trying to build the fake registrar XML).

spec/support/authentication_helpers.rb
```diff
@@ -59,4 +49,29 @@ def password_from_config_or_prompt
...
+  end
+
+  def ensure_password!
+    self.password ||= password_from_config_or_prompt
+  end
+
+  def submit_credentials_if_needed
+    return unless page.has_text?('SUNet ID', wait: Settings.post_authentication_text_timeout)
+
+    ensure_password!
```
jmartin-sul, line 64: as far as i could tell, password (unlike sunet) is only used for login, so it seemed fine to keep it coupled to login form submission.  happy to adjust as needed if testing by others indicates that this doesn't work (this is the main reason i'd like someone who doesn't use cardinal key to test).

spec/features/etd_creation_spec.rb
```diff
@@ -207,12 +207,13 @@
...
       end
     end
     expect(page).to have_text(dissertation_title)
     apo_element = find_table_cell_following(header_text: 'Admin policy')
-    expect(apo_element.first('a')[:href]).to have_text('druid:bx911tp9024') # this is hardcoded in hydra_etd app
+    expect(apo_element.first('a')[:href]).to include('druid:bx911tp9024') # this is hardcoded in hydra_etd app
```
jmartin-sul, line 216: i'm not sure how this passed for other folks before?  when the tests got here for me with the old code, it complained that `apo_element.first('a')[:href]` was a `String`, and so the `has_text?` method was unavailable.  e.g.
```ruby
expected "https://argo-stage.stanford.edu/view/druid:bx911tp9024" to respond to `has_text?`
```

mjgiarlo: I wound up having to change this in my PR too.

ndushay (commented):
h2 test passed;

etd test hung due to slowness uploading files  :-(

that's not new to this PR, though.  New since I was last FR, but ... I'll worry about it next week.

I have cardinalKey on my box;  i don't use it with regular firefox;  not sure if firefox browser instance used for int tests uses it or not.

jmartin-sul:
> h2 test passed;
> 

nice!

> etd test hung due to slowness uploading files :-(
> 

😞 

> that's not new to this PR, though. New since I was last FR, but ... I'll worry about it next week.
> 

yeah, that's been the flakiest spec for me on this FR shift.  has failed numerous times, and passed maybe 2 or 3 times.

> I have cardinalKey on my box; i don't use it with regular firefox; not sure if firefox browser instance used for int tests uses it or not.

did you get prompted for a password or see the user/pass login form on this run?

mjgiarlo:
I think a few of us may want to pair on this tomorrow, if folks are interested and available.

spec/features/etd_creation_spec.rb
jmartin-sul: i prefer what #517 does to deal with this problem

jmartin-sul:
superseded by #522

PR #524, opened 2022-11-18
Upgrades sdr client and fixes tests.

justinlittman:
Get the correct sdr-api client version.

### DIFF
--- a/spec/support/deposit_helpers.rb
+++ b/spec/support/deposit_helpers.rb
@@ -32,7 +32,6 @@ module DepositHelpers
         filenames.first => { 'preserve' => true, 'publish' => false, 'shelve' => false },
         filenames.last => { 'preserve' => false, 'publish' => true, 'shelve' => true }
       }
-      filenames = filenames.map { |file| "spec/fixtures/#{file}" }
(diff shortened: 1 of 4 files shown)


spec/support/deposit_helpers.rb
```diff
@@ -32,7 +32,6 @@ def deposit_object(filenames: [])
         filenames.first => { 'preserve' => true, 'publish' => false, 'shelve' => false },
         filenames.last => { 'preserve' => false, 'publish' => true, 'shelve' => true }
       }
-      filenames = filenames.map { |file| "spec/fixtures/#{file}" }
       file_set_type_strategy = SdrClient::Deposit::ImageFileSetStrategy
```
peetucket, line 35: is this removed because it was never used?

jmartin-sul (approved):

spec/support/deposit_helpers.rb
jmartin-sul: fwiw, i didn't have a great sense of the specific motivation for the change, so i ran `spec/features/sdr_deposit_spec.rb` before i merged this (passed for me)

PR #562, opened 2023-03-15
add new spec for registering objects with folio_instance_hrid

### LINKED ISSUE #561: Register/refresh with HRID spec
Create a new test that:
1. Registers in Argo with an HRID.
2. Waits for registration to complete.
3. Check the facets.
4. Checks the title on the show page.
5. Changes the HRID.
6. Refreshes the metadata.
7. Checks the title on the show page.

peetucket:

Fixes #561 

Add a new spec to verify we can register objects in Argo using a folio_instance_hrid and it pulls the label from Folio.

Note: This will only work if the environment being tested has folio enabled across all apps.  This is currently only true in QA.  This is determined via the setting.

Yes

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -45,5 +45,7 @@ gis:
   robots_content_root: '/var/geomdtk/current/stage'
   username: 'lyberadmin'
 
+test_folio_instance_hrid: 'a10065784'
+
 folio:
   enabled: false
--- a/config/settings/qa.yml
+++ b/config/settings/qa.yml
@@ -18,3 +18,6 @@ was_registrar:
   jobs_directory: '/was_unaccessioned_data/jobs'
   url: 'https://was-registrar-app-qa.stanford.edu'
   username: 'was'
+
+folio:
+  enabled: true
--- a/spec/features/create_folio_object_spec.rb
+++ b/spec/features/create_folio_object_spec.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Use Argo to create an item object with a folio instance HRID' do
+  let(:random_word) { random_phrase }
+  let(:object_label) { "Object Label for #{random_word}" }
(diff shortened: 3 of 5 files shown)


config/settings.yml
```diff
@@ -45,5 +45,7 @@ gis:
   robots_content_root: '/var/geomdtk/current/stage'
   username: 'lyberadmin'
 
+test_folio_instance_hrid: 'a10065784'
+
```
peetucket, line 49: could possibly be different per environment

spec/features/create_folio_object_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    click_link 'Refresh'
+    reload_page_until_timeout!(text: catalog_object_label_updated) # updated label pulled from folio for new HRID
+
+    # look for metadata source facet
+    visit Settings.argo_url
+    click_link 'Show more facets'
+    click_button 'Metadata Source (Multi)'
+    within('.blacklight-metadata_source_ssim') do
+      expect(page).to have_text('Folio')
+    end
```
peetucket, line 72: it would be nice to check the total number of objects shown in this facet and see if it increased by one after we registered this object, but I can't think of a good way of finding this number (there isn't an obvious selector to use)

justinlittman: Check out the `access_indexing_spec` for testing indexing via a facet.

peetucket: Ok, so basically search for the single druid and see if the relevant facet value is 1.  Thanks.  I replaced this part of the test with that check.

config/settings/qa.yml
```diff
@@ -18,3 +18,6 @@ was_registrar:
   jobs_directory: '/was_unaccessioned_data/jobs'
   url: 'https://was-registrar-app-qa.stanford.edu'
   username: 'was'
+
+folio:
+  enabled: true
```
peetucket, line 23: folio is enabled in the QA environment currently

PR #566, opened 2023-03-19
reset the shelve workflow step if it fails

### LINKED ISSUE #556: Re-run accessionWF shelve step if it errors
`h2_object_creation_spec.rb` often fails when the following error occurs in the accessionWF:

`Error: shelve : problem with shelve (BackgroundJob: 468659): {:errors=>[{:title=>"Unable to shelve files", :detail=>"Unable to find sul-logo.png in the content directory"}]}`".

If you manually re-run the shelve step in the accessionWF via Argo, the workflow completes successfully. 

This ticket is to add re-running the shelve step to the integration test in the case of the error occurring.

peetucket:

Fixes #556 - the shelve step often fails in the H2 test (possible race condition).  This resets it automatically if it does.

Localhost -- success several times in a row (watching the step fail and then succeed)


spec/features/h2_object_creation_spec.rb
```diff
@@ -113,12 +113,21 @@
...
+    visit "#{Settings.argo_url}/view/#{bare_druid}"
+    Timeout.timeout(Settings.timeouts.workflow) do
+      loop do
+        break if page.has_text?('v2 Accessioned', wait: 1)
+
+        reset_errored_workflow_step('accessionWF') if page.has_text?('Error: shelve : problem with shelve', wait: 0)
+        sleep 3 # wait to ensure the workflow step reset was accepted so we don't reset again immediately
+        page.refresh
+      end
+    end
```
peetucket, line 127: this borrows but modifies some logic from the page helper method `reload_page_until_timeout!` to look for the shelve step in error, and then calls out to a new method to reset the error

spec/support/workflow_helpers.rb
```diff
@@ -0,0 +1,14 @@
...
+  # resets the first errored out step in the given workflow
+  def reset_errored_workflow_step(workflow_name)
+    click_link workflow_name
+    select 'Rerun', from: 'status'
+    confirm_message = 'You have selected to manually change the status. '
+    confirm_message += 'This could result in processing errors. Are you sure you want to proceed?'
+    accept_confirm(confirm_message) do
+      click_button 'Save'
+    end
+  end
```
peetucket, line 13: Note: due to no good HTML selectors, it's not easy to properly select a specific step in the workflow for resetting... so this assumes there is a single step in error, and it will just reset that one (we can at least limit ourselves to a specific workflow)

edsu (approved):
I left a comment, but it's not a blocker I guess.

jmartin-sul (approved):

mjgiarlo (approved):

PR #569, opened 2023-03-20
[SPIKE] Prevent SDR env defaulting from leaking all over the place

mjgiarlo:

* Prevent SDR env defaulting from leaking all over the place
  * Use a new global variable to reflect which SDR environment the tests are running against
    * Rubocop doesn't like global variables but I'm not sure we have a better option
  * Inlines config gem setup in `spec_helper.rb` to ensure it happens before anything else, given how much relies on the config gem
  * Move Capybara setup into a dedicated file
    * Clearer delineation between RSpec setup and Capybara setup
* Make spec helpers more self-contained
  * This removes one step from adding or removing helpers, so the spec helper no longer needs direct knowledge of which support helpers need including.

N/A


peetucket:
Was thinking the same thing as I added another SDR_ENV check to the code.  Looks good to me. Have you run the suite in QA and stage to be sure it still runs as expected?

mjgiarlo:
@peetucket 💬 
> Was thinking the same thing as I added another SDR_ENV check to the code. Looks good to me. Have you run the suite in QA and stage to be sure it still runs as expected?

I did, and it looked OK to me, but I thought it'd be good for someone else to try this too given how idiosyncratic the tests can be.

peetucket:
Ok, will run in stage and QA and report back.

peetucket (approved):
All pass on stage, can't get ETD one to pass on QA, likely due to environment setup though.  H2 also has the shelve issue in both cases... I think because you didn't have the latest PR in before you branched.

mjgiarlo:
@peetucket Aha, very likely. Thank you!

PR #580, opened 2023-04-04
update Settings.goobi.url to point to newly built goobi stage server

jmartin-sul:

goobi stage was rebuilt to use ubuntu as its OS, and now lives at a slightly different hostname

n/a

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -37,7 +37,7 @@ sunet:
 
 # the URL and username/password used to login to goobi (applicable to stage only)
 goobi:
-  url: https://goobi-stage.stanford.edu/goobi
+  url: https://goobi-stage-a.stanford.edu/goobi
   username: ~
   password: ~
 


ndushay (approved):

config/settings.yml
```diff
@@ -37,7 +37,7 @@ sunet:
 
 # the URL and username/password used to login to goobi (applicable to stage only)
 goobi:
-  url: https://goobi-stage.stanford.edu/goobi
+  url: https://goobi-stage-a.stanford.edu/goobi
   username: ~
```
peetucket, line 41: this will be the new permanent name (even if later, that old name becomes an alias to the new name, right)?  in other words, this config won't have to change back again?

jmartin-sul: that's my understanding, yeah.

@rsmith11 was suggesting aliasing the old hostname to this new VM, once the prod cutover is done.  at which point we may want to revert to the hostname that this PR moves away from.  but for some (maybe short) period of time, it sounds like `goobi-stage.stanford.edu` won't be a usable hostname, and for now it's the deprecated old VM anyway.

and also yeah, i think goobi-stage-a will be the "real" hostname that goobi-stage may point to later, but goobi-stage-a should still work fine even after that aliasing.  so switching back shouldn't be urgent, if i'm understanding correctly.

PR #582, opened 2023-04-10
PageHelpers.reload_page_until_timeout_with_wf_step_retry! method added

ndushay:

in PR #566, Peter added code to rerun a workflow step if needed.  This makes that approach into something clearer (if verbose) and puts it to use in an additional test.   Both tests using this new method failed for me consistently at the same workflow steps.

### DIFF
--- a/spec/features/h2_object_creation_spec.rb
+++ b/spec/features/h2_object_creation_spec.rb
@@ -116,24 +116,11 @@ RSpec.describe 'Use H2 to create a collection and an item object belonging to it
 
     # Opens Argo detail page
     visit "#{Settings.argo_url}/view/#{bare_druid}"
-    reload_page_until_timeout!(check_wf_error: false) do
-      # NOTE: This is here to work around a persistent, not easily
-      #       reproducible race condition that is occasionally seen in Argo,
-      #       causing integration tests to fail. This work-around mimics the
-      #       steps that developers tend to perform manually.
-      if page.has_text?('Error: shelve : problem with shelve', wait: 0)
-        click_link 'accessionWF'
-        select 'Rerun', from: 'status'
-        confirm_message = 'You have selected to manually change the status. '
-        confirm_message += 'This could result in processing errors. Are you sure you want to proceed?'
-        accept_confirm(confirm_message) do
-          click_button 'Save'
-        end
-        sleep 3
-      end
-
-      page.has_text?('v2 Accessioned', wait: 3)
-    end
+    # wait for accessioningWF to finish; retry if error on shelving step, likely caused by a race condition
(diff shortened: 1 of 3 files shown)


spec/features/h2_object_creation_spec.rb
```diff
@@ -116,24 +116,11 @@
...
-        sleep 3
-      end
-
-      page.has_text?('v2 Accessioned', wait: 3)
-    end
+    # wait for accessioningWF to finish; retry if error on shelving step, likely caused by a race condition
+    reload_page_until_timeout_with_wf_step_retry!(expected_text: 'v2 Accessioned',
+                                                  workflow: 'accessionWF',
+                                                  workflow_retry_text: 'Error: shelve : problem with shelve',
+                                                  retry_wait: 10)
```
ndushay, line 123: this is the same functionality but more re-usable and clearer (IMO)

ndushay: Note that this needs a much longer wait after the retry of the step than the other test using this method.

spec/features/item_creation_no_files_spec.rb
```diff
@@ -49,8 +49,11 @@
     expect(page).to have_text("Project : #{project}")
     expect(page).to have_text("Registered By : #{AuthenticationHelpers.username}")
 
-    # wait for accessioningWF to finish
-    reload_page_until_timeout!(text: 'v1 Accessioned')
+    # wait for accessioningWF to finish; retry if Version mismatch on sdr-ingest-transfer
+    reload_page_until_timeout_with_wf_step_retry!(expected_text: 'v1 Accessioned',
+                                                  workflow: 'accessionWF',
+                                                  workflow_retry_text: 'Version mismatch',
+                                                  retry_wait: 2)
```
ndushay, line 56: the addition of the retry here and below made this test pass much more consistently.

spec/support/page_helpers.rb
```diff
@@ -1,14 +1,40 @@
...
 
         # Check for workflow errors and bail out early. There is no recovering
         # from a workflow error. This selector is found on the Argo item page.
-        expect(page).not_to have_css('.alert-danger', wait: 0) if check_wf_error
+        expect(page).not_to have_css('.alert-danger', wait: 0)
+
+        page.refresh
+      end
+    end
+  end
```
ndushay, line 16: for this method, I just reverted the changes in PR #566 because I split that functionality into a separate method.

peetucket (approved):
LGTM.  I had started down this approach as well, but we decided since it was only one spot to not do this originally.  But it makes sense if used in other tests.

PR #587, opened 2023-04-26
Change permissions of manually created WARC file

mjgiarlo:

Connects to sul-dlss/was_robot_suite#583

`Tempfile.new` defaults to 0600 permissions (user r/w) and this breaks the web archive accessioning spec when we deploy presbots with the tarpipe-to-cp change. This commit lets the test pass with that new code deployed.


edsu (approved):
Did we recently switch from tarpipe-to-cp?

mjgiarlo:
@edsu There's been an on-hold PR for switching from tarpipe to cp. Now it's merged and deployed everywhere.

Thanks for the review!

PR #590, opened 2023-05-11
Upgrade to Ruby 3.2.2

mjgiarlo:

Keep pace with Ruby updates

Yes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,3 +6,6 @@ workflows:
     jobs:
       - ruby-rails/lint:
           name: lint
+          executor:
+            name: ruby-rails/ruby
+            ruby-tag: '3.2.2'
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -6,3 +6,6 @@ workflows:
     jobs:
       - ruby-rails/lint:
           name: lint
+          executor:
+            name: ruby-rails/ruby
+            ruby-tag: '3.2.2'
```
ndushay, line 11: are you thinking of making a new release of our circleci orb for this?

ndushay (approved):

.circleci/config.yml
mjgiarlo: @ndushay Yep. And then removing all the manual CI ruby stuff.

PR #591, opened 2023-05-12
Bump sleeps in pre-assembly specs to allow them to pass

mjgiarlo:

To help the pre-assembly tests pass in stage. They were consistently failing with the prior sleep values, now that Ruby 3.2 is deployed everywhere. They were the only tests that failed.

I'm not sure how much to be concerned about this, given that Andrew's testing in stage has not reproduced this error condition. He's tried with a batch of small files and a batch of large files, and neither wound up in this situation.

No.

### DIFF
--- a/spec/features/preassembly_hfs_accessioning_spec.rb
+++ b/spec/features/preassembly_hfs_accessioning_spec.rb
@@ -106,7 +106,7 @@ RSpec.describe 'Create and re-accession object with hierarchical files via Pre-a
 
     expect(find_table_cell_following(header_text: 'Content type').text).to eq('file') # filled in by accessioning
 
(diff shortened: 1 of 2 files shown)


peetucket (approved):

peetucket:
Seems odd to be connected to ruby 3.2, but not odd enough to spend time worrying about.  Once the sleeps starting getting really long, maybe we can worry more.

mjgiarlo:
@peetucket 💬 
> Seems odd to be connected to ruby 3.2, but not odd enough to spend time worrying about. Once the sleeps starting getting really long, maybe we can worry more.

I feel the same way. If we care a little bit, we could toss an issue at this repo to figure out why we even need them and put that in the low priority area on the FR board? I'm not sure I do care that much 😆

PR #596, opened 2023-05-30
explicitly add websocket as a gem

ndushay:

I couldn't run the integration tests when I upgraded to ruby 3.2.2;  they failed with this error:

```
$ bx rspec

An error occurred while loading spec_helper. - Did you mean?
                    rspec ./spec/spec_helper.rb

Failure/Error: require 'selenium-webdriver'

LoadError:
  cannot load such file -- websocket
# ./spec/spec_helper.rb:11:in `<top (required)>
```

I could run rspec with ruby 3.1.2, but not with ruby 3.2.2.   when I added the `websocket` gem on the command line for ruby 3.2.2, I could run rspec with ruby 3.2.2.   Hence, my adding it to the Gemfile.  It may or may not fix the problem for the next person upgrading to ruby 3.2.2.

I also note https://github.com/titusfortner/webdrivers/blob/main/README.md?plain=1#L8-L12, which may imply we won't need the webdrivers gem much longer (?)  ... which is not to say we wouldn't need websockets.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,3 +23,4 @@ gem 'rubyXL' # for updating Excel spreadsheets
 gem 'sdr-client', '~> 2.0'
 gem 'selenium-webdriver'
 gem 'webdrivers'
+gem 'websocket' # required by selenium-webdrivers; a ruby upgrade somehow obscured this dependency
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -231,6 +231,7 @@ GEM
 PLATFORMS
   ruby
   x86_64-darwin-19
+  x86_64-darwin-22
   x86_64-linux
 
 DEPENDENCIES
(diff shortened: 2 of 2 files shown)


Gemfile
```diff
@@ -23,3 +23,4 @@ gem 'rubyXL' # for updating Excel spreadsheets
 gem 'sdr-client', '~> 2.0'
 gem 'selenium-webdriver'
 gem 'webdrivers'
+gem 'websocket' # required by selenium-webdrivers; a ruby upgrade somehow obscured this dependency
```
ndushay, line 26: I know this secondary dependency _shouldn't_ be necessary, but perhaps it will surface the problem better, even if it doesn't solve it.

justinlittman (approved):

PR #603, opened 2023-07-11
fix label text used to find Metadata Source facet button in test for Argo item creation with folio HRID

jmartin-sul:
looks like the relevant argo change was made here: https://github.com/sul-dlss/argo/commit/36bacbe475d45f73c791816d20ffbf6578e08352

keep the test suite up to date.  discovered when running against QA after deploying this week's dep updates.

nope


mjgiarlo (approved):
Good catch!

PR #610, opened 2023-08-12
adding comments pertaining to folio cutover

ndushay:

to decouple comment language from symphony

it wasn't necessary.


spec/features/etd_creation_spec.rb
```diff
@@ -235,7 +235,7 @@
     else
       click_link('etdSubmitWF')
       within('#blacklight-modal') do
-        # expect first 5 steps to have completed w/ Symphony
+        # expect first 5 steps to have completed w/ ILS catalog
```
justinlittman, line 238: Again. would suggest "Folio" instead of "ILS catalog".

justinlittman (approved):

PR #625, opened 2023-09-08
etd spec waits for submit-marc and check-marc steps to complete

### LINKED ISSUE #619: etd test needs to wait for submit-marc and catalog-status steps to complete
ETD test fails when it is checking Argo workflow modal for folio steps that take a bit to run.

Ideally, ETD test would have a loop that:

- opens ETD workflow and looks for all steps through check-marc completed
   - if completed, we're done, break out of loop and move on
   - if not completed, cancel out of the workflow modal, then load it again
 
Dirty fix is just to wait long enough before checking

ndushay:

Fixes #619

For me, on a fast fiber connection, checking the ETD workflow in argo was happening before the submit-marc and check-marc etdSubmitWF steps were completed, so the test would fail.   This PR fixes that problem.

no need.

### DIFF
--- a/spec/features/etd_creation_spec.rb
+++ b/spec/features/etd_creation_spec.rb
@@ -216,40 +216,31 @@ RSpec.describe 'Create a new ETD with embargo, and then update the embargo date'
     expect(apo_element.first('a')[:href]).to end_with('druid:bx911tp9024') # this is hardcoded in hydra_etd app
     status_element = find_table_cell_following(header_text: 'Status')
     expect(status_element).to have_text('v1 Registered')
-    if Settings.folio.enabled
-      reload_page_until_timeout! do
+
+    Timeout.timeout(Settings.timeouts.workflow) do
+      workflow_done = false # a "break" statement inside the "within" block does not break out of the loop
+      loop do
         click_link('etdSubmitWF')
-        # expect first 6 steps to have completed w/ Folio (but requires reload since the Folio job takes time)
-        page.has_text?(/register-object completed/) &&
-          page.has_text?(/submit completed/) &&
-          page.has_text?(/reader-approval completed/) &&
-          page.has_text?(/registrar-approval completed/) &&
-          page.has_text?(/submit-marc completed/) &&
-          page.has_text?(/check-marc completed/)
+        within('.modal-dialog') do
+          workflow_done = true if page.has_text?(/register-object\s+completed/) &&
+                                  page.has_text?(/submit\s+completed/) &&
+                                  page.has_text?(/reader-approval\s+completed/) &&
+                                  page.has_text?(/registrar-approval\s+completed/) &&
+                                  page.has_text?(/submit-marc\s+completed/, wait: 1) &&
+                                  page.has_text?(/check-marc\s+completed/, wait: 1) &&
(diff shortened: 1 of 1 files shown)


spec/features/etd_creation_spec.rb
```diff
@@ -216,40 +216,32 @@
     expect(apo_element.first('a')[:href]).to end_with('druid:bx911tp9024') # this is hardcoded in hydra_etd app
     status_element = find_table_cell_following(header_text: 'Status')
     expect(status_element).to have_text('v1 Registered')
-    if Settings.folio.enabled
```
ndushay, line 219: since I was already making changes right here, I just yanked the folio switch out.

spec/features/etd_creation_spec.rb
```diff
@@ -216,40 +216,32 @@
...
-        expect(page).to have_text(/reader-approval completed/)
-        expect(page).to have_text(/registrar-approval completed/)
-        expect(page).to have_text(/submit-marc completed/)
-        expect(page).to have_text(/check-marc waiting/)
-
-        # NOTE: it would be lovely if we could process the ETD through the rest of the etdSubmitWF steps
-        #   and then run it through common-accessioning, but the remaining etdSubmitWF steps of check-marc,
-        #   catalog-status, and otherMetadata require too much fakery specific to the ETD app (cocina-model
-        #   updates) to make sense here.
+        within('.modal-dialog') do
```
ndushay, line 224: because this is a timeout loop that has to look at a modal, it didn't seem worth it to generalize this situation.  Perhaps that will change in the future.  For now, I made it bespoke to ETDs.

spec/features/etd_creation_spec.rb
```diff
@@ -216,40 +216,32 @@
...
+            page.has_text?(/submit-marc\s+completed/, wait: 1) &&
+            page.has_text?(/check-marc\s+completed/, wait: 1) &&
+            page.has_text?(/catalog-status\s+waiting/, wait: 1)
+
+          # NOTE: it would be lovely if we could process the ETD through the rest of the etdSubmitWF steps
+          #   and then run it through common-accessioning, but the remaining etdSubmitWF steps, catalog-status and
+          #   otherMetadata, require too much fakery specific to the ETD app (cron job, cocina-model updates)
+          #   to make sense here.
+
+          page.send_keys(:escape)
```
ndushay, line 239: it turns out that "Cancel" is not unique on the page and we don't have identifiers or other unique selectors on these blacklight modal buttons.  So I opted to go with the escape key to close the modal.

spec/features/etd_creation_spec.rb
```diff
@@ -216,40 +216,32 @@
     expect(apo_element.first('a')[:href]).to end_with('druid:bx911tp9024') # this is hardcoded in hydra_etd app
     status_element = find_table_cell_following(header_text: 'Status')
     expect(status_element).to have_text('v1 Registered')
-    if Settings.folio.enabled
-      reload_page_until_timeout! do
+
+    Timeout.timeout(Settings.timeouts.workflow) do
+      workflow_done = false
```
ndushay, line 221: this is clunky, but apparently "break" only gets out of the "within" and we needed to get out of both the "within" and the "loop" statement.

mjgiarlo (approved):

PR #644, opened 2023-10-10
Test shelve, publish, and preserve.

### LINKED ISSUE #42: test variants of shelve / publish / preserve settings in accessioning
these settings are how we do things across systems;  it would be appropriate to have tests to indicate the right things happen for these settings.

we should have tests ensuring that
- preserve = yes puts a file in preservation
- preserve = no does NOT put a file in preservation
- shelve = yes puts the file in stacks (or clearly does the right trigger for getting the file in stacks)
- shelve = no does NOT put the file in stacks
- publish = yes puts the file in purl (or clearly does the right trigger for getting the file in purl) 
- publish = no does NOT put the file in purl

When an object with files with different settings has gone through assembly/common accessioning, viewing the object in argo shows the files and these attributes at the bottom of the page

Given sul-dlss/dor-services-app/4617, this ticket is a "go"

justinlittman:
closes #42

### DIFF
--- a/spec/features/preassembly_hfs_accessioning_spec.rb
+++ b/spec/features/preassembly_hfs_accessioning_spec.rb
@@ -157,11 +157,8 @@ RSpec.describe 'Create and re-accession object with hierarchical files via Pre-a
     expect_text_on_purl_page(druid:, text: collection_name)
 
     # verify the cocina json has the filenames with paths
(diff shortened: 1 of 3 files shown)


mjgiarlo (approved):

ndushay:
❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

PR #663, opened 2024-01-09
revert some click_link_or_button changes back to click_button

jmartin-sul:

The former is ambiguous for argo object registration and search.

I like the spirit of the rubocop suggestion, but for an integration test of other existing codebases, it seems a bit presumptuous.  So, I went with inline exceptions and a comment, since it might be nice to have rubocop nudge us towards better page element disambiguation when writing new app code (which seems like the driver for most, though not all, of our new integration tests?).  But I'm totally happy to just disable the rubocop rule with an explanation in `.rubocop.yml`, and get rid of the inline comments and exceptions.

Also, for the cases that make a difference in Argo, I'm guessing that this isn't a huge deal: I doubt that sighted users will confuse the "Register" nav link at the top of the page with the "Register" button that submits the form that the aforementioned link takes you to.  And I'm guessing (though grain of salt since this isn't an area of expertise) that visually impaired users who are using screen readers will be able to easily tell from the context and experience with other pages that the nav link and the button have different purposes.  Maybe the "Search" thing will be trickier, harder for me to judge that, since the "Search" links that got confused with the "Search" button didn't seem to be visible to me in the Argo show page.

n/a

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -487,8 +487,6 @@ Style/ReturnNilInPredicateMethodDefinition: # new in 1.53
   Enabled: true
 Style/YAMLFileRead: # new in 1.53
   Enabled: true
-Capybara/ClickLinkOrButtonStyle: # new in 2.19
-  Enabled: true
 Capybara/RSpec/HaveSelector: # new in 2.19
   Enabled: true
 Capybara/RSpec/PredicateMatcher: # new in 2.19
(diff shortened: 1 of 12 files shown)


justinlittman:
That seems like a lot of additional and redundant comments in the code. I'd vote for just disabling the rule.

mjgiarlo:
UGH! I see now that the fix I put in for this is sitting locally. I never pushed it or made a PR. I'm sorry about that.

Instead of disabling the rule, an alternative is to set `EnforcedStyle: strict # more backwards-compatible` in the rubocop config for this rule.

jmartin-sul:
> UGH! I see now that the fix I put in for this is sitting locally. I never pushed it or made a PR. I'm sorry about that.
> 
> Instead of disabling the rule, an alternative is to set `EnforcedStyle: strict # more backwards-compatible` in the rubocop config for this rule.

ah, ok, and no worries!  want me to switch to that in this PR, or would you rather just push your branch and have me close this?  i'm happy to go either way.

jmartin-sul:
i would also say that i have no personal preference between disabling the rule, and using `EnforcedStyle: strict`.  if i'm reading the rubocop docs correctly, the former will flag the use of `click_link_or_button`.  which some might consider a downside?  but having been unaware of that method before digging into this issue, i don't think it's something i'd reach for anyway.  though i can see the benefit of preferring to reach for  `click_link_or_button` first.

mjgiarlo:
@jmartin-sul Doesn't matter much to me either. But this PR looks shippable so let's run with it!

mjgiarlo (approved):

PR #665, opened 2024-01-11
H2: click through the terms of deposit modal when it's presented

jmartin-sul:

I was getting presented with a terms of deposit modal in the H2 tests modified here, and since the tests didn't handle it, I had to babysit them and click through manually to get them to pass.  This PR should hopefully alleviate the babysitting.

_**Example of the modal:**_
Screenshot 2024-01-09 at 11 49 31 AM

Here's an example screencast of me successfully letting the two modified tests run.  It's a bit long, but you can see the terms modal flit by at 0:10 and 0:25 for one test; and 1:04 and 1:46 for the other.  It was 65 MB over github's 100 MB limit, hence this drive link: https://drive.google.com/file/d/1n2EOWYXZykIJJZ2x1FoydiivouSQyNiv/view?usp=drive_link

_**screenshot of the test output:**_

Screenshot 2024-01-11 at 1 19 18 PM

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -3,6 +3,7 @@ timeouts:  # in seconds
   workflow: 300
   bulk_action: 200
   post_authentication_text: 5
+  h2_terms_modal_wait: 5 # should show quickly, so can be much shorter than default wait time
   events:
     poll_for: 240
     poll_interval: 2
--- a/spec/features/h2_globus_creation_spec.rb
+++ b/spec/features/h2_globus_creation_spec.rb
@@ -29,6 +29,7 @@ RSpec.describe 'Use H2 to create a collection and an item object belonging to it
     select 'CC0-1.0', from: 'collection_required_license'
 
     click_link_or_button 'Deposit'
+    click_through_terms_of_deposit_modal
     expect(page).to have_text(collection_title)
     expect(page).to have_text('+ Deposit to this collection')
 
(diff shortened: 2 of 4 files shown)


mjgiarlo (commented):
Given each use of the new method immediately follows `click_link_or_button 'Deposit'`, I wonder if the helper changes could include doing that step as well. That way, if/when we add more H2 tests, we don't need to remember "when I click the Deposit button, I need to also remember to add the line about the terms modal" or something?

jmartin-sul:
> Given each use of the new method immediately follows `click_link_or_button 'Deposit'`, I wonder if the helper changes could include doing that step as well. That way, if/when we add more H2 tests, we don't need to remember "when I click the Deposit button, I need to also remember to add the line about the terms modal" or something?

ah yeah, i had considered that briefly, though hadn't given it much thought.  my rationale for not doing that was that it might not be any more obvious that the special all-in-one helper should get used instead of just using `click_link_or_button 'Deposit'`.  but that wasn't something i felt strongly about, so if you think that'd be more usable, happy to go that way instead.

whichever approach is used, a comment at the top of each H2 spec file might be helpful?

mjgiarlo:
@jmartin-sul Sounds good!

jmartin-sul:
> @jmartin-sul Sounds good!

@mjgiarlo i tacked on a small refactor commit that takes your suggestion and adds a comment warning our future selves to use the helper method.

jmartin-sul:
(i also re-ran `h2_object_creation_spec.rb` just to make sure i didn't break anything, and i saw it click through the modal with the refactored code)

mjgiarlo (approved):

mjgiarlo:
@jmartin-sul thx! Looks great.

PR #667, opened 2024-01-15
revert bulk_tags_edit_spec change

ndushay:

To fix a mistake I added to dep updates.

### DIFF
--- a/spec/features/bulk_tags_edit_spec.rb
+++ b/spec/features/bulk_tags_edit_spec.rb
@@ -57,7 +57,8 @@ RSpec.describe 'Use Argo to edit administrative tags in bulk' do
           wait_for_download
           druids_with_tags = CSV.parse(File.read(download))
           expect(druids_with_tags.count).to eq(number_of_druids)
-          druids_with_tags.each do |druid, tags|
+          # druids_with_tags is an array of arrays, each of which is a druid and its tags
+          druids_with_tags.each do |(druid, *tags)|
             expect(druid).to match(druid_regex)
             tags.all? { |tag| expect(tag).to match(tag_regex) }
           end


justinlittman (approved):

spec/features/bulk_tags_edit_spec.rb
```diff
@@ -57,7 +57,7 @@
           wait_for_download
           druids_with_tags = CSV.parse(File.read(download))
           expect(druids_with_tags.count).to eq(number_of_druids)
-          druids_with_tags.each do |druid, tags|
+          druids_with_tags.each do |druid, *tags|
```
jmartin-sul, line 60: the rubocop exception comment needs to be restored too, i think: https://github.com/sul-dlss/infrastructure-integration-test/pull/666/commits/c4edb8e7123ab38cbeab016d5f28ef37b470907c#diff-1413f9d516eee302bb7e789fe56d0135bf7edc823b4dfbf04b63c5ece6188511L60

```suggestion
          druids_with_tags.each do |druid, *tags| # rubocop:disable Style/HashEachMethods rubocop seems mistaken in thinking the tags var is unused
```

jmartin-sul (changes requested):
i think the rubocop exception comment needs to be re-added to keep the lint check from failing.

ndushay:
@jmartin-sul after a bunch of digging, I figured out what was happening in rubocop and changed the syntax used so the rubocop exception is no longer needed.

image

jmartin-sul:
> @jmartin-sul after a bunch of digging, I figured out what was happening in rubocop and changed the syntax used so the rubocop exception is no longer needed

nice, thanks for the explanation!

jmartin-sul (approved):

PR #669, opened 2024-01-18
remove unnecessary code now that bug is fixed

peetucket:

We do not need this manual completion anymore now that https://github.com/sul-dlss/gis-robot-suite/issues/401 is fixed

See https://github.com/sul-dlss/workflow-server-rails/pull/707

Tested with this fixed deployed to stage

Nope


edsu (approved):
🎉 yay!

PR #675, opened 2024-01-24
Remove manual start of accessionWF for gis test

lwrubel:
Tests that accessionWF is kicked off from gisDeliveryWF. 

N/A


edsu (approved):
Simplification! 🥳

PR #676, opened 2024-01-24
Create new gis preassembly spec

peetucket:

https://github.com/sul-dlss/pre-assembly/pull/1403 allows for GIS objects to be accessioned via pre-assembly.  This adds a new integration test for this pathway since it exercises some different pre-assembly code.  Note that it places the GIS test data to be accessioned in a server location just as we do for the other pre-assembly specs.

This requires a new config setting, and while doing this, I added a commit which refactors some of the settings/configs to do three things:
- make the preassembly settings block more like the others (nested)
- move settings that are exactly the same between qa/stage into the settings.yml file to avoid duplication
- removed the need/config to SSH into the GIS robot server (the test data needed for GIS tests is now mounted on pre-assembly servers, and we are already configured to SSH into those servers for moving data around)

I also updated the backup gis integration data ZIP file to better match what is actually sent in for real objects

All tests were run on both stage and QA with these settings/config refactors and everything passes.

Nope.


config/settings.yml
```diff
@@ -44,6 +44,12 @@ goobi:
 
 gis:
   robots_content_root: '/gis_workflow_data/stage'
-  username: 'lyberadmin'
```
peetucket, line 47: do not need this anymore, the GIS test data can now be found on the pre-assembly server because of the new mount, so one less server to SSH into

config/settings.yml
```diff
@@ -44,6 +44,12 @@ goobi:
...
 
 test_folio_instance_hrid: 'a10065784'
+
+# note for purl_url: there is no purl for qa;  sul-h2-qa uses stage ...
+purl_url: 'https://sul-purl-stage.stanford.edu' # from shared_configs for sul-h2-stage
+preassembly:
+  username: 'preassembly'
+  bundle_directory: '/dor/staging/integration-tests/files-reaccessioning-test'
+  hfs_bundle_directory: '/dor/staging/integration-tests/hierarchical-file-test'
```
peetucket, line 55: 1. these are all the same between qa/stage, put them here instead
2. nest the preassembly related items in a block instead of using `preassembly_bundle_directory` etc

config/settings/qa.yml
```diff
@@ -3,14 +3,9 @@ dor_services:
   url: 'https://dor-services-qa-lb.stanford.edu'
 etd_url: 'https://etd-qa.stanford.edu'
 h2_url: 'https://sul-h2-qa.stanford.edu'
-gis:
-  robots_host: 'kurma-robots-qa-01.stanford.edu'
-# for purl_url: there is no purl for qa;  sul-h2-qa uses stage ...
```
peetucket, line 8: do not need to ssh into the GIS server anymore, since test data is accessible via the already configured pre-assembly servers

config/settings/qa.yml
```diff
@@ -3,14 +3,9 @@ dor_services:
...
-purl_url: 'https://sul-purl-stage.stanford.edu'
-preassembly_bundle_directory: '/dor/staging/integration-tests/files-reaccessioning-test'
-preassembly_hfs_bundle_directory: '/dor/staging/integration-tests/hierarchical-file-test'
-preassembly_host: 'sul-preassembly-qa.stanford.edu'
-preassembly_url: 'https://sul-preassembly-qa.stanford.edu'
+preassembly:
```
peetucket, line 6: moved to `settings.yml` as described above

spec/features/gis_accessioning_spec.rb
```diff
@@ -31,16 +31,16 @@
     # wait for object to be registered
     expect(page).to have_text 'Items successfully registered.'
 
-    bare_object_druid = find('table a').text
-    puts " *** GIS accessioning druid: #{bare_object_druid} ***" # useful for debugging
-    druid = "druid:#{bare_object_druid}"
+    bare_druid = find('table a').text
+    puts " *** GIS accessioning druid: #{bare_druid} ***" # useful for debugging
+    druid = "druid:#{bare_druid}"
```
peetucket, line 37: we use `bare_druid` in other tests, just make this match to reduce confusion

spec/features/gis_accessioning_spec.rb
```diff
@@ -31,16 +31,16 @@
...
 
     # Go to kurma server and copy test content to the druid folder so it can be accessioned
     # Should this test data be deleted from the server,
     # a zipped copy is available at spec/fixtures/gis_integration_test_data.zip
     test_data_source_folder = File.join(Settings.gis.robots_content_root, 'integration_test_data')
-    test_data_destination_folder = File.join(Settings.gis.robots_content_root, bare_object_druid, 'temp')
-    copy_command = "ssh #{Settings.gis.username}@#{Settings.gis.robots_host} " \
+    test_data_destination_folder = File.join(Settings.gis.robots_content_root, bare_druid, 'temp')
+    copy_command = "ssh #{Settings.preassembly.username}@#{Settings.preassembly.host} " \
                    "\"mkdir -p #{test_data_destination_folder} " \
```
peetucket, line 44: GIS test data is accessible on pre-assembly servers now

aaron-collier (approved):

PR #679, opened 2024-01-30
[HOLD] update expected filenames for GIS

peetucket:

Goes with https://github.com/sul-dlss/gis-robot-suite/pull/742

If we rename the data.zip and normalized ZIP filenames in the gis robots, we need to update the integration tests to look for them.

[HOLD] until/if https://github.com/sul-dlss/gis-robot-suite/pull/742 is merged

Verified test pass with new robot change deployed to stage.


peetucket:
Not needed since we are not proceeding with renaming of zips.

PR #714, opened 2024-04-09
Improve reliability of integration tests

mjgiarlo:
# Why was this change made?

This commit helped the tests that were touched pass more regularly for me.

Includes:
* Stop using the timeout + loop strategy with a janky modal-closing routine
  * It's no longer needed; the workflow modal updates dynamically via Turbo
* Extend H2 "continue deposit" timeout
* Make sure Argo registration form has fully loaded (including turbo frames) before submitting
* Remove noisy `puts` output from H2 helpers

# Was README.md updated if necessary?

No


justinlittman (approved):
All 4 tests passed on first try.

mjgiarlo:
@justinlittman Thanks for giving this a spin!

PR #722, opened 2024-04-25
Switch to the redesigned sdr-client interface

mjgiarlo:
# Why was this change made?

Connects to sul-dlss/sdr-client#261

Ran all tests in stage and they all passed.

### DIFF
--- a/spec/features/h2_globus_creation_spec.rb
+++ b/spec/features/h2_globus_creation_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe 'Use H2 to create a collection and an item object belonging to it
   # may pop up on any attempt to deposit.
   scenario do
     # remove modal for deposit in progress, if present, waiting a bit for some rendering
-    click_link_or_button 'No' if page.has_text?('Continue your deposit')
+    click_link_or_button 'No' if page.has_text?('Continue your deposit', wait: Settings.timeouts.h2_terms_modal_wait)
 
     # CREATE COLLECTION
     click_link_or_button '+ Create a new collection'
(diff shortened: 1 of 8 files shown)


spec/features/h2_globus_creation_spec.rb
```diff
@@ -16,7 +16,7 @@
   # may pop up on any attempt to deposit.
   scenario do
     # remove modal for deposit in progress, if present, waiting a bit for some rendering
-    click_link_or_button 'No' if page.has_text?('Continue your deposit')
+    click_link_or_button 'No' if page.has_text?('Continue your deposit', wait: Settings.timeouts.h2_terms_modal_wait)
```
mjgiarlo, line 19: I removed these a week or two ago, but I find that I need them to make sure I can get past the "Continue your deposit" modal.

spec/features/sdr_deposit_spec.rb
```diff
@@ -14,16 +14,15 @@
   end
 
   it 'deposits objects' do
-    ensure_token
```
mjgiarlo, line 17: No need to sprinkle the token code all over since it's handled within the client now.

spec/support/authentication_helpers.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module AuthenticationHelpers
-  mattr_accessor :username, :password, :token
+  mattr_accessor :username, :password
```
mjgiarlo, line 4: The token is only used for SDR API operations, so move that code into the deposit helpers.

jcoyne (approved):

PR #752, opened 2024-06-24
Add require csv where necessary

aaron-collier:

Add `require 'csv'` in the places that we open a CSV now that the gem is explicitly included.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -2,6 +2,7 @@
 
 require 'byebug'
 require 'config'
+require 'csv'
 require 'faker'
 require 'io/console'
 require 'pry-byebug'


mjgiarlo (commented):
It looks like I mistakenly removed the `require` statement from `spec/spec_helper.rb` in the flurry of csv-related PRs late last week. I think we can safely re-add the `require` line to that file which I removed here: https://github.com/sul-dlss/infrastructure-integration-test/commit/ea5c06efda6fbbc6634b39a11f691719df516922#diff-89eebfcbc0f14b6d989517837ca1e94fce4e2ce9a03233641cd936f2b8d2ed94 That way we don't need to remember to manually require if/when we need CSV functionality next time.

mjgiarlo (approved):

PR #759, opened 2024-07-22
Test OCR

edsu:

OCR functionality is now in main and should be tested in Stage and QA
even if it hasn't been deployed yet to production.


lwrubel:
i think you also need to change/remove the ocr settings in `settings/stage.yml`  to not override this.

lwrubel (approved):

PR #762, opened 2024-07-24
Adds H2 versioned test.

### LINKED ISSUE #761: New H2 user version integration test
Initially this test should:
1. Deposit a version.
2. Make sure that the user version is indicated on Argo details page.
3. Deposit a new version, not creating a new user version.
4. Make sure that the user version is indicated for the correct object version on Argo details page.
5. Deposit a new version, create a new user version.
6. Make sure that the new user version is indicated for the correct object version on Argo details page.

Note that user versions is only enabled for H2 QA.

justinlittman:
closes #761

Test user versioning from H2.

### DIFF
--- a/spec/features/h2_versioned_object_creation_spec.rb
+++ b/spec/features/h2_versioned_object_creation_spec.rb
@@ -1,0 +1,141 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Use H2 to create a collection and a versioned work belonging to it' do
+  let(:collection_title) { random_phrase }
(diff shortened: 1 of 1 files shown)


mjgiarlo (approved):
I'm pondering how different this is from the existing H2 test, and if we want two end-to-end H2 tests. Approving, though, since I would be surprised if you hadn't already considered this.

justinlittman:
Yes, this test focuses specifically on user version related behavior. The other H2 test is more comprehensive, e.g., testing embargoes and reserving PURLs.

PR #790, opened 2024-10-17
speech to text integration test

### LINKED ISSUE #787: Write integration test for speech to text WF
Verify the speechToTextWF works as expected for audio and video via accessioning from pre-assembly through to purl, analogus to the ocr specs for images and document.

Useful to have early on so we can easily push objects through the workflow and find where it's still broken

peetucket:

Fixes #787 

For accessioning a media object and sending it through speechToTextWF

Useful for testing purposes.

HOLD until the the implementation is complete and we add more expectations to the files produced.


config/settings.yml
```diff
@@ -54,10 +54,14 @@ preassembly:
...
   hfs_bundle_directory: '/dor/staging/integration-tests/hierarchical-file-test'
   gis_bundle_directory: '/dor/staging/integration-tests/gis-test'
 
 earthworks_url: 'https://earthworks-stage.stanford.edu/catalog'
 
 ocr:
   enabled: true
+
+speech_to_text:
+  enabled: false
```
peetucket, line 67: you can override this to `true` in `config/settings/qa.local.yml` or `config/settings/stage.local.yml` to run the test on your laptop when you are confident the correct branches of pre-asembly and common-accessioning are deployed

eventually will become true for everyone when in production

spec/features/preassembly_ocr_image_spec.rb
```diff
@@ -26,12 +25,6 @@
       #{bare_druid},content
     CSV
   end
-  let(:preassembly_reaccession_manifest_csv) do
-    <<~CSV
-      druid,object
-      #{bare_druid},#{bare_druid}
-    CSV
-  end
```
peetucket, line 34: unused

spec/support/page_helpers.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 module PageHelpers
-  def reload_page_until_timeout!(text: '')
-    Timeout.timeout(Settings.timeouts.workflow) do
+  def reload_page_until_timeout!(text: '', num_seconds: Settings.timeouts.workflow)
+    Timeout.timeout(num_seconds) do
```
peetucket, line 5: the text to speech process takes a good while, this let's us up the timeout setting for just the one spot that takes a long time, leaving the default configured time everyone else

edsu (approved):

edsu (approved):

PR #794, opened 2024-11-04
Allow configuration of firefox/geckodriver paths and disable pop-up download panel

mjgiarlo:
# Why was this change made?

I needed it to deal with how Firefox is being shipped in modern Ubuntu versions. I don't think it'll break anyone else since it's opt-in, but it may be good to verify that.

# Was README.md updated if necessary?

No.

### DIFF
--- a/spec/support/capybara.rb
+++ b/spec/support/capybara.rb
@@ -8,8 +8,11 @@ Capybara.enable_aria_label = true
 Capybara.run_server = false
 
 Capybara.register_driver :my_firefox_driver do |app|
+  Selenium::WebDriver::Firefox::Service::EXECUTABLE = Settings.browser.geckodriver_path if Settings.browser.geckodriver_path
+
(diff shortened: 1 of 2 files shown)


spec/support/capybara.rb
```diff
@@ -8,8 +8,11 @@
 Capybara.run_server = false
 
 Capybara.register_driver :my_firefox_driver do |app|
+  Selenium::WebDriver::Firefox::Service::EXECUTABLE = Settings.browser.geckodriver_path if Settings.browser.geckodriver_path
+
   options = Selenium::WebDriver::Firefox::Options.new
   options.profile = Selenium::WebDriver::Firefox::Profile.new.tap do |profile|
+    profile['browser.download.alwaysOpenPanel'] = false
```
mjgiarlo, line 15: This annoys the heck out of me every single time I run the integration tests.

justinlittman (approved):

PR #795, opened 2024-11-05
retry ocrWF fetch-files if preservation-client retrieval errors

jmartin-sul:

this accounts for a race condition where fetch-files fails to get the files on the first try or 3, because there seems to occasionally be some lag before the files are available for read by the preservation_catalog web VMs.  previously, the spec failed on line 119 due to the workflow error.

plus an opportunistic YARD typo fix.

example of successful accessioning after manually retrying (this just has the test do the same automatically).

note: i think this is ok, and is something we've seen intermittently on stage and not worried about, including with ocrWF, but i'm not 100% sure about that.  if it feels like this is papering over a serious problem, we can work with ops to investigate the lag (probably a good idea even if this fix is also a good idea, and could reproduce using the code from before this change).

discovered incidentally when testing https://github.com/sul-dlss/dor-services-app/pull/5218

Screenshot 2024-11-04 at 11 15 00 PM

Screenshot 2024-11-04 at 11 15 57 PM

Screenshot 2024-11-04 at 11 16 52 PM

n/a

### DIFF
--- a/spec/features/preassembly_ocr_image_spec.rb
+++ b/spec/features/preassembly_ocr_image_spec.rb
@@ -116,7 +116,10 @@ RSpec.describe 'Create an image object via Pre-assembly and ask for it be OCRed'
     reload_page_until_timeout!(text: 'ocrWF')
 
     # Wait for the second version accessioningWF to finish
-    reload_page_until_timeout!(text: 'v2 Accessioned')
+    workflow_retry_text = /fetch-files : Preservation::Client.*got.*from Preservation/
+    reload_page_until_timeout_with_wf_step_retry!(expected_text: 'v2 Accessioned',
+                                                  workflow: 'ocrWF',
+                                                  workflow_retry_text:)
 
     # Check that the version description is correct for the second version
     reload_page_until_timeout!(text: 'Start OCR workflow')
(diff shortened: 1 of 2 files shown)


edsu (commented):
I thought this wasn't really needed after the try/catch logic around retrying was adjusted to catch errors when streaming to S3?

https://github.com/sul-dlss/common-accessioning/pull/1396 ?

edsu:
Ooops, I didn't mean to close this!

edsu (approved):
Approving, but I thought this wasn't really needed after the try/catch logic around retrying was adjusted to catch errors when streaming to S3?

https://github.com/sul-dlss/common-accessioning/pull/1396 ?

peetucket (approved):

peetucket:
oops, i merged and then read the comment thread - sorry, can revert if we'd like

jmartin-sul:
> oops, i merged and then read the comment thread - sorry, can revert if we'd like

oh, right, i still need to do the testing described above.  i think it's harmless to leave in.  i have a TODO in my notes to do the above described testing, if that works out an eliminates the need for this change, i'll put up a PR to revert this.

PR #801, opened 2024-12-04
preserve json files for stt

peetucket:

Goes with https://github.com/sul-dlss/common-accessioning/pull/1431

This updates the media test to handle the json files being part of the object


peetucket:
closed in favor of #804

PR #803, opened 2024-12-09
Prefer using binstubs to explicit `bundle exec`

mjgiarlo:
# Why was this change made?

- **Prefer binstubs to `bundle exec`**
- **Keep warnings out of test output**


peetucket (approved):

edsu (approved):
Just curious, why is using binstubs preferable to bundle exec?

mjgiarlo:
@edsu 💬 
> Just curious, why is using binstubs preferable to bundle exec?

It's kind of a six vs half-dozen situation, but:

1. Less typing, both when running the command and when documenting commands in READMEs and such;
2. Less `rbenv` smell (as an `rvm`-with-gemsets user, I haven't typed `bundle exec` in years);
3. Recommended within `rbenv`-land (apparently?): https://github.com/rbenv/rbenv/wiki/understanding-binstubs
4. Faster (arguably?): https://andre.arko.net/2021/02/22/never-bundle-exec-again/

edsu:
Thanks! All those sound good to me. The more that they are uniform across projects the better. I honestly never know quite what to do and always default to using `bundle exec`.

PR #817, opened 2025-01-23
Increase speechToText timeout

edsu:

It takes about 9 minutes for AWS Batch to provision the EC2 instance and then for the job to complete.

Testing in QA


jmartin-sul (approved):

edsu:
@peetucket  I guess we could only do one video instead of two? Also we could cut the video in half or a 1/3...

peetucket:
> @peetucket I guess we could only do one video instead of two? Also we could cut the video in half or a 1/3...

yeah, finding shorting videos may be the long term solution --- i don't think we need to worry too much for now 

the reason for two was that one is an audio and the other is a video (even though it's the same content) to be sure both media types work

peetucket:
Also, the AWS spin up time may be a limiting factor in how fast the tests run even if we had a shorter video, so may not be much we can do about that

PR #822, opened 2025-01-30
require csv for ruby 3.4.1

peetucket:

need to explicitly require csv when running on ruby 3.4 or later (csv gem is already included in gemfile)


mjgiarlo (approved):
Good catch!

PR #829, opened 2025-02-20
Error handling improvements

jmartin-sul:

file system lag (?) made `sdr_deposit_spec` almost unusable for me otherwise this week.

since the console output prints the info about the screenshot fairly prominently, and since the screenshotting isn't configurable in this version of the PR, it didn't seem necessary to adjust the readme.  but i'm of course happy to add to the readme for this PR as-is if that's desired, and/or to make the screenshotting behavior configurable (e.g. enabled via setting, with whatever default people prefer).

### DIFF
--- a/spec/features/sdr_deposit_spec.rb
+++ b/spec/features/sdr_deposit_spec.rb
@@ -56,6 +56,8 @@ RSpec.describe 'SDR deposit' do
     # Wait for accessioningWF to finish
     reload_page_until_timeout!(text: 'v1 Accessioned')
 
+    # We think this is needed due to network file system lag (between write to disk finishing, and
+    # initial visibility to client VMs).  waiting briefly usually works, so retry for a bit.
     retries_count = 0
     begin
       # Download Gemfile (preserved=true) from Preservation
@@ -72,7 +74,6 @@ RSpec.describe 'SDR deposit' do
         raise 'Error opening opening file modal'
       end
     rescue StandardError => e
-      # we think this happens due to network file system lag.  retrying usually works.
       puts "download attempt failed (link click text=#{gemfile_pres_link_text}): #{e.class}; #{e.inspect}; #{e}"
 
       puts "sleeping and retrying (retries_count=#{retries_count})"
@@ -87,6 +88,8 @@ RSpec.describe 'SDR deposit' do
 
     click_link_or_button 'Cancel'
 
+    # We think this is needed due to network file system lag (between write to disk finishing, and
+    # initial visibility to client VMs).  waiting briefly usually works, so retry for a bit.
     retries_count = 0
     begin
       # Download Gemfile.lock (shelve=true) from Stacks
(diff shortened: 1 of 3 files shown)


spec/features/sdr_deposit_spec.rb
```diff
@@ -56,26 +56,64 @@
...
 
-    # Download Gemfile (preserved=true) from Preservation
-    click_link_or_button 'Gemfile'
-    expect(page).to have_text 'Preservation:'
-    gemfile_pres_link = find('.modal-content a')
-    gemfile_pres_url = gemfile_pres_link['href']
-    expect(gemfile_pres_url.end_with?("/items/#{druid}/files/Gemfile/preserved?version=1")).to be true
-
-    click_link_or_button gemfile_pres_link.text
+    retries_count = 0
```
justinlittman, line 59: Can you put in a comment about why you're retrying?

spec/features/sdr_deposit_spec.rb
```diff
@@ -56,26 +56,64 @@
...
 
-    # Download Gemfile.lock (shelve=true) from Stacks
-    click_link_or_button 'Gemfile.lock'
-    expect(page).to have_text 'Stacks:'
-    gemfile_lock_stacks_link = find('.modal-content a')
-    gemfile_lock_stacks_url = gemfile_lock_stacks_link['href']
-    expect(gemfile_lock_stacks_url.end_with?("/file/#{druid}/Gemfile.lock")).to be true
-
-    click_link_or_button gemfile_lock_stacks_link.text
+    retries_count = 0
```
justinlittman, line 90: Same.

spec/spec_helper.rb
```diff
@@ -60,6 +60,23 @@
     mocks.verify_partial_doubles = true
   end
 
+  # cribbed from https://gist.github.com/osulyanov/10609515 and https://rspec.info/documentation/3.13/rspec-core/RSpec/Core/Example.html
```
justinlittman, line 63: Can you put in a comment that says what this is doing?

spec/features/sdr_deposit_spec.rb
jmartin-sul: i did put one at the top of the `rescue` block, but i'm happy to move it up here if you think that's a better location (i think maybe it is, now that you've pointed this out)

spec/features/sdr_deposit_spec.rb
jmartin-sul: same

spec/spec_helper.rb
jmartin-sul: yup, good call

jmartin-sul:
@justinlittman i threw on a commit that puts the retry comments in more useful locations, adds a comment about desired behavior to the screenshot config block, and does a little other touchup i noticed while there.  merging since you approved, but happy to circle back if that doesn't cover it or you noticed anything else.

PR #837, opened 2025-03-10
Update H3 feature test

mjgiarlo:
NOTE: It still does not pass due to some bugs we are actively working on, but it is better now than it was before.

### DIFF
--- a/spec/features/h3_object_creation_spec.rb
+++ b/spec/features/h3_object_creation_spec.rb
@@ -121,7 +121,7 @@ RSpec.describe 'Use H3 to create a collection and an item object belonging to it
     find('.nav-link', text: 'Deposit', exact_text: true).click
 
     # if you have ever agreed to the terms, there will be no checkbox
-    check('I agree to the SDR Terms of Deposit') if page.has_css?('#work_agree_to_terms', wait: 0)
+    check('I agree to the SDR Terms of Deposit') if page.has_css?('#work_agree_to_terms')
 
     click_link_or_button 'Deposit', class: 'btn-primary'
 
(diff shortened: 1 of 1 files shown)


spec/features/h3_object_creation_spec.rb
```diff
@@ -118,10 +119,11 @@
     end
 
     find('.nav-link', text: 'Deposit', exact_text: true).click
-    click_link_or_button 'Deposit', class: 'btn-primary'
 
-    #   # if you have previously agreed to the terms within the last year, there will be no checkbox
-    #   check('I agree to the SDR Terms of Deposit') if page.has_css?('#work_agree_to_terms', wait: 0)
+    # if you have previously agreed to the terms within the last year, there will be no checkbox
```
justinlittman, line 123: `# if you have ever agreed to the terms ...`

spec/features/h3_object_creation_spec.rb
```diff
@@ -118,10 +119,11 @@
...
+    check('I agree to the SDR Terms of Deposit') if page.has_css?('#work_agree_to_terms', wait: 0)
```
justinlittman, line 124: Is the `wait: 0` necessary?

justinlittman (approved):
Approve pending consideration of comments.

spec/features/h3_object_creation_spec.rb
mjgiarlo:
```suggestion
    # if you have ever agreed to the terms, there will be no checkbox
```

spec/features/h3_object_creation_spec.rb
mjgiarlo:
```suggestion
    check('I agree to the SDR Terms of Deposit') if page.has_css?('#work_agree_to_terms')
```

PR #845, opened 2025-03-31
Update the H3 integration test for updated label text

aaron-collier:

This passes for me (woot) although slow and prone to timeout with argo, but this should pass for most.


mjgiarlo (approved):

peetucket:
thanks - i also verified it worked for me.

PR #889, opened 2025-07-29
[SPIKE] Switch to Playwright over Selenium and update initial test

aaron-collier:

This is an initial spike to investigate if switching to [Playwright](https://github.com/YusukeIwaki/capybara-playwright-driver) over Selenium will improve the consistency of our tests when running as a suite, and possibly the performance.

Articles of note:
- https://justin.searls.co/posts/running-rails-system-tests-with-playwright-instead-of-selenium/


spec/features/gis_accessioning_spec.rb
```diff
@@ -62,8 +62,12 @@
                                                   workflow: 'gisAssemblyWF',
                                                   workflow_retry_text: 'Error: extract-boundingbox',
                                                   retry_wait: 5) do |page|
-      click_link_or_button 'Reindex'
+      page.driver.with_playwright_page do |page|
```
aaron-collier, line 65: There seems to be a known issue (https://github.com/YusukeIwaki/capybara-playwright-driver/issues/83) where playwright just doesn't find something in the DOM. This line consistently raised a `not found in DOM` error when testing. The "workaround" described in the linked issue does work, which might lend itself to adding a helper method if there are a sufficient number of these situations.

spec/features/gis_accessioning_spec.rb
```diff
@@ -83,15 +87,53 @@
...
-    expect(files[0].text).to match(%r{AirMonitoringStations.shp application/vnd.shp 8.14 KB})
-    expect(files[1].text).to match(%r{AirMonitoringStations.shx application/vnd.shx 2.39 KB})
-    expect(files[2].text).to match(%r{AirMonitoringStations.dbf application/vnd.dbf 40.8 KB})
-    expect(files[3].text).to match(%r{AirMonitoringStations.prj text/plain 468 Bytes})
-    expect(files[4].text).to match(%r{preview.jpg image/jpeg 2\d.\d KB})
-    expect(files[5].text).to match(%r{AirMonitoringStations.shp.xml application/xml 6\d.\d KB})
-    expect(files[6].text).to match(%r{AirMonitoringStations-iso19139.xml application/xml 2\d.\d KB})
-    expect(files[7].text).to match(%r{AirMonitoringStations-iso19110.xml application/xml 1\d.\d KB})
-    expect(files[8].text).to match(%r{AirMonitoringStations-fgdc.xml application/xml 5.\d+ KB})
+
```
aaron-collier, line 90: This was an interesting situation. Playwright did not match the text across table cells the way this was originally written. For instance I got:
```
- expected: 'AirMonitoringStations.shp\t\t\tapplication/vnd.shp 8.14 KB'
- got: 'AirMonitoringStations.shp\t\t\tapplication/vnd.shp 8.14 KB'
```
with some variation on the whitespace between the tabs.

By instead verifying the cells and then the individual text, this seems a bit clearer and performant (but that could just be my perception)

README.md
```diff
@@ -13,6 +13,14 @@ The tests depend on having Firefox (default) or Chrome downloaded.
 1. `bundle install`
 1. `bin/rake webdrivers:geckodriver:update`
 
+### Install the Playwright CLI
+
+```
+export PLAYWRIGHT_CLI_VERSION=$(bundle exec ruby -e 'require "playwright"; puts Playwright::COMPATIBLE_PLAYWRIGHT_VERSION.strip')
+yarn add -D "playwright@$PLAYWRIGHT_CLI_VERSION"
+yarn run playwright install
```
mjgiarlo, line 21: Is this a one-time thing or should we be running it after we `bundle install` (in case the `playwright` gem bumps)?

PR #913, opened 2025-10-06
remove cdl; deal with file system latency

peetucket:

1. CDL is turned off, comment out that part of the test
2. File system latency is causing another test to fail sporadically, this uses a strategy similar to that used in another test to wait for stacks to be "ready"


spec/features/access_indexing_spec.rb
```diff
@@ -63,9 +63,10 @@
     # choose_rights('Location: Music Library (no-download)')
     # find_access_rights_single_facet_value(object_druid, 'location: music (no-download)')
 
+    # October 2025: CDL turned off, so test commented out.
```
justinlittman, line 66: CDL is being deleted from the codebase. Please delete here.

justinlittman (approved):
Approve pending handling of comment.

PR #985, opened 2026-04-21
[DRAFT | HOLD] Methodically assess integration tests until deterministic pass/fail

aaron-collier:

This draft PR is largely to capture systematic changes, test by test to try to get to a point where pass/fail is deterministic.

Methodology:
- Remove all tests and start with just one: `sdr_client_deposit_spec`
- Individually add tests back in and get it to pass consistently if possible.
- Surface difficult to resolve bugs when spread across the suite (particularly lead to cascading failures)
- Each commit to this branch will focus on a single test being reintroduced.

### AI Assistance:

Part of the approach to this work has been capturing the test output between comparable runs and using AI to analyze the output in order to surface bugs:

```
bundle exec rspec --format documentation 2>&1 | tee tmp/run_1.txt
```
Capture the seed value:

```
bundle exec rspec --seed NNNN --format documentation 2>&1 | tee tmp/run_2.txt
```

This has proven particularly helpful when duplicate seed runs fail in different ways.

### DIFF
--- a/spec/features/sdr_client_deposit_spec.rb
+++ b/spec/features/sdr_client_deposit_spec.rb
@@ -41,7 +41,13 @@ RSpec.describe 'SDR client deposit to SDR API' do
 
     # Tests existence of technical metadata
     expect(page).to have_text 'Technical metadata'
-    click_link_or_button 'Technical metadata'
+    # Scroll to the bottom so the lazily-loaded tech metadata section enters the viewport
+    # and the browser fetches its content.
+    page.scroll_to(:bottom)
+    button = find_button('Technical metadata')
+    execute_script('arguments[0].scrollIntoView(true)', button)
+    button.click
+    # click_link_or_button 'Technical metadata'
     sleep(5)
 
     # Scroll to the bottom so the lazily-loaded tech metadata section enters the viewport
@@ -49,9 +55,17 @@ RSpec.describe 'SDR client deposit to SDR API' do
     page.scroll_to(:bottom)
 
     within('#document-techmd-section') do
-      file_listing = find_all('.file')
-      # Only preserved files get techmd
-      expect(file_listing.size).to eq 2
+      retries_count = 0
+      begin
+        file_listing = find_all('.file')
+        # Only preserved files get techmd
+        expect(file_listing.size).to eq 2
+      rescue StandardError
+        retries_count += 1
+        puts "No files found with techmd, waiting and retrying (#{retries_count}/5)..."
+        sleep 5
+        retry if retries_count < 5
+      end
     end
 
     # Wait for accessioningWF to finish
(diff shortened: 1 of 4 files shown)


spec/features/sdr_client_deposit_spec.rb
```diff
@@ -41,17 +41,33 @@
 
     # Tests existence of technical metadata
     expect(page).to have_text 'Technical metadata'
-    click_link_or_button 'Technical metadata'
+    # Scroll to the bottom so the lazily-loaded tech metadata section enters the viewport
+    # and the browser fetches its content.
+    page.scroll_to(:bottom)
```
aaron-collier, line 46: There is a known bug/issue with the underlying automation technology in browsers (Marionette for Firefox) when scrolling and finding objects on the page. For Firefox it often will fail to interact with an element that is found on the page (so the expect on like 43 passes) but it's below the existing window and firefox doesn't scroll to it. An explicit `execute_script` is more reliable. Here I'm also including an additional `page.scroll_to(:bottom)` to aid in getting things on screen.

spec/features/sdr_client_deposit_spec.rb
```diff
@@ -41,17 +41,33 @@
...
-      # Only preserved files get techmd
-      expect(file_listing.size).to eq 2
+      retries_count = 0
+      begin
+        file_listing = find_all('.file')
+        # Only preserved files get techmd
+        expect(file_listing.size).to eq 2
+      rescue StandardError => e
+        retries_count += 1
+        puts "No files found with techmd, waiting and retrying (#{retries_count}/5)..."
```
aaron-collier, line 65: There is often a delay in techmd being available on the page, retry for a bit.

spec/support/download_helpers.rb
```diff
@@ -22,7 +22,7 @@ def download_content
   def wait_for_download
     # raising StandardError with a helpful message makes failure reporting nicer
     Timeout.timeout(TIMEOUT, StandardError, 'timed out waiting for download') do
-      sleep 0.1 until downloaded?
+      sleep 5 until downloaded?
```
aaron-collier, line 25: this download sleep has proven to be too slow. `5` seems to work reliably for me, but it may need to be a bit longer.

spec/support/xlsx_helpers.rb
```diff
@@ -4,7 +4,7 @@
 
 module XlsxHelpers
   def create_druid
-    source_id = "spreadsheet-druid:#{random_phrase}"
+    source_id = "spreadsheet-druid:#{SecureRandom.uuid}"
```
aaron-collier, line 7: `random_phrase`, provided by Faker, is seeded in such a way that the "randomness" is somewhat unreliable. `SecureRandom.uuid` is much more random, so appending it to `source_id` should avoid random "already exists" failures that end up cascading.

PR #986, opened 2026-04-22
Split tests between "submit", "accession", and "review" to improve reliability of specs

aaron-collier:
As a review of this PR may be a bit confusing, a few points:
- I don't believe any test functionality (with the exception of virtual objects, and a chunk of the preassembly re-accessioning spec that I'm working through at this time) has been lost.
- This is largely just shuffling code around. The part of most tests that start with using argo to register an object is moved into a shared example and a single initial test to register all objects first.
- The accessioning and sdr specs maintain the sections of tests that interact with the objects.
- The verify specs maintain the final sections of tests that check the result of accessioning.

## Summary

This is an experiment toward establishing a more reliable approach to infrastructure integration testing so that we feel more confidence in the results.

One of the most problematic sources of failure in the integration tests is waiting for content complete the accessioning process in order for the expected results to be displayed. The cascading nature of these failures has to do with waiting on the various services to complete their workflows and update the object, thus refreshing the page to get that content is necessary but often results in false failures from either not properly reloading the DOM, stale DOM state, or known bugs when content is not initially present on the page.

The goal of this separation experiment is to determine if structuring the order of tests to flow through the logical states of SDR objects (register -> accession -> review) improves that reliability and open a discussion about the goals of our integration testing and provide a possible pattern that can either be finalized and adopted or used to identify individual test improvements that we can incrementally adopt.

## Observations

A few notes about the structure of PR and why some choices were made.

- Use custom type metadata to tag each tests with the logical step in an SDR objects lifecycle (`registration`, `accessioning`, `verify`) and two additional groups (`sdr` and  `preassembly`).
- The folder structure under `spec/features/` based on the metadata type is ONLY in place for now as a means of offering an easier review. There is no functional reason to follow this organization unless we decide to going forward.
- Forcing the ordering of tests in `spec_helper.rb`:
  - Lines 137-168 implements the use of the custom `type` metadata in order to set the order of tests by group as `registration`, then `accessioning`, then `sdr`, then `verify`, then `preassembly` with transitional pauses between groups.
  - Line 100 excludes `preassembly` tests from running unless manually included (i.e. `bin/rspec -t type:preassembly`) -- Note, that `preassembly` must be defined in the order above in order to ever be included, but this line keeps them from running as part of `bin/rspec` (reasoning discussed later).
- Establish `shared_examples` in `spec/support` to address extensive duplication, like using argo to register objects (which we do around 15 times) or setting up preassembly jobs.
- For lack of a better name, the `sdr` tests cover the tests that fall outside of the argo registration or preassembly workflow. For now, that's H3 objects, ETD, and bulk tags editing. The ordering of that group after `accessioning` and before `verify` is intentional to provide an arbitrary delay before verifying that accessioning tasks complete.

### Time spent in accessioning is somewhat unpredictable

This is the reason that most of our tests currently use the `reload_page_until_timeout!`, but often fail either before the timeout or result in a stale DOM and never get the refreshed content. By separating many of our tests at those calls avoids the DOM problem but doesn't entirely solve the timing issue without sufficient time between the tests. While this structure has been far more reliable for me, it is not foolproof.

### Preassembly, oh preassembly

The preassembly tests are, for me, the most unreliable. There are several reasons for this, but I suspect the main issues are the delays introduced by the slower services that these test touch (OCR, GIS, and Folio) mean that some or most of the `reload_page_until_timeout` calls hit their time outs. This causes the entirety of the infrastructure test suite to run very slowely, and also seems to be the cause of selenium related bugs or failures that cascade into all tests.

TDB


spec/spec_helper.rb
```diff
@@ -129,7 +129,28 @@
...
+  # config.order = :random
+  config.register_ordering(:global) do |examples|
+    submit_examples = examples.select { |ex| ex.file_path.include?('/features/submit/') }
+    review_examples = examples.select { |ex| ex.file_path.include?('/features/review/') }
+    middle_examples = examples - submit_examples - review_examples
+
+    # Tag the last submit example as the boundary
+    last_submit = submit_examples.last
+    last_submit.metadata[:submit_boundary] = true if last_submit
+
```
aaron-collier, line 145: This tags the last submission test as the boundary to force the sleep below on line 151

spec/spec_helper.rb
```diff
@@ -129,7 +129,28 @@
...
+    # Randomize within each bucket so intra-group order dependencies are still surfaced
+    submit_examples.shuffle + middle_examples.shuffle + review_examples.shuffle
```
aaron-collier: This is forcing all submission tests to run first, anything not in submission or review to run next, then the review specs to run last. This should allow background jobs to complete prior to the review specs so we can try to avoid and cleanup the peppering of `sleeps` and `waits` in individual tests.

spec/spec_helper.rb
```diff
@@ -129,7 +129,28 @@
...
+  end
+
+  # Determine if the submit boundary has been reached and pause after it
+  # This allows background processing of submitted examples to complete before the review phase starts.
+  config.after(:each) do |example| # rubocop:disable RSpec/HookArgument
+    if example.metadata[:submit_boundary]
+      puts "\nSubmit phase complete — pausing 60 seconds..."
+      sleep 60
```
aaron-collier: I chose 60 seconds here for 2 reasons:
- With only the sdr_client_deposit spec, the object is accessioned in around 20 seconds (can be verified in argo by looking at the event timestamps)
- However, if files are going to end up on stacks, that can take a bit longer, 60 seconds was the most reliable pause I could come up with.

Given how much we currently pause and wait for content in existing tests, 1+ minute of pausing here between tests seems inconsequential.

mjgiarlo (approved):

PR #1002, opened 2026-05-28
preassembly_reaccessioning_spec.rb confirms replicated copy is retrievable and passes checksum validation

jmartin-sul:
uses preservation_catalog's `bin/fixity_check_replicated_moabs` script to confirm cloud archive is retrievable (from at least one endpoint) and complete/correct (_NOTE_: depends on https://github.com/sul-dlss/preservation_catalog/pull/2530)

More complete automated end to end testing.

Shouldn't add much time since it's one test, and the Moab being retrieved and validated is small.

n/a

### DIFF
--- a/spec/features/preassembly_reaccessioning_spec.rb
+++ b/spec/features/preassembly_reaccessioning_spec.rb
@@ -296,5 +296,6 @@ RSpec.describe 'Create and re-accession image object via Pre-assembly' do
 
     visit_argo_and_confirm_event_display!(druid:, version: latest_version)
     confirm_archive_zip_replication_events!(druid:, from_version: 1, to_version: latest_version)
+    retrieve_and_fixity_check_replicated_druid!(bare_druid:, expected_version: latest_version)
   end
 end
--- a/spec/support/preservation_helpers.rb
+++ b/spec/support/preservation_helpers.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module PreservationHelpers
+  def retrieve_and_fixity_check_replicated_druid!(bare_druid:, expected_version:)
+    fixity_check_cmd = 'cd preservation_catalog/current && RAILS_ENV=production bin/fixity_check_replicated_moabs ' \
+                       "--druid_list #{bare_druid} --endpoints_to_audit gcp_s3_south_1"
+    remote_destination = "#{Settings.preservation_catalog.username}@#{Settings.preservation_catalog.host}"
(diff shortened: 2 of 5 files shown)


spec/features/preassembly_reaccessioning_spec.rb
```diff
@@ -296,5 +296,20 @@
...
+        Open3.capture3(remote_fixity_check_cmd)
+      rescue StandardError => e
+        puts "Error executing system command: '#{remote_fixity_check_cmd}' raised #{e}"
+      end
+
+    cmd_result = { cmd: remote_fixity_check_cmd, stdout_str:, stderr_str:, exitstatus: status.exitstatus, success: status.success? }
+    puts cmd_result
+    expect(cmd_result[:exitstatus]).to eq 0
+    expect(cmd_result[:success]).to be true
+    expect(stdout_str).to match(/fixity check passed - validate_checksums - #{bare_druid}.*actual version: #{latest_version}/)
```
jmartin-sul: i'll likely pull this out into a helper method before taking this PR out of draft

spec/support/preservation_helpers.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module PreservationHelpers
+  def retrieve_and_fixity_check_replicated_druid!(bare_druid:, expected_version:)
+    fixity_check_cmd = 'cd preservation_catalog/current && RAILS_ENV=production bin/fixity_check_replicated_moabs ' \
+                       "--druid_list #{bare_druid} --endpoints_to_audit gcp_s3_south_1"
+    remote_destination = "#{Settings.preservation_catalog.username}@#{Settings.preservation_catalog.host}"
+    remote_fixity_check_cmd = "ssh #{remote_destination} '#{fixity_check_cmd}'"
+    stdout_str, stderr_str, status =
```
jmartin-sul, line 9: i tested on stage with a moab i purposely corrupted, as well as with a regular valid moab.  and in both cases, the exit code from the bin script was passed back from the remote host and returned by `ssh` (whether the exit code was zero or not).

(i tested the audit failure case by using the cloud fixity check script on a random small moab, deleting files from the downloaded zips after that passed, then re-running the cloud fixity check on the same druid, which let it use the corrupted cached zips instead of re-downloading fresh valid zips.  the exit code of 1 when errors were detected was passed back from the fixity check script run on the remote host.)

mjgiarlo (approved):

PR #1003, opened 2026-06-01
[SPIKE] add basic Word Error Rate check to preassembly_speech_to_text_media_spec.rb

jmartin-sul:
- Introduce `amatch` and `webvtt-ruby` dependencies to support transcription quality assessment.
- Implement a word-to-character mapping strategy in `TranscriptionHelpers` to allow the character-based Amatch C extension to perform accurate word-level Levenshtein distance calculations (1 word change = 1 edit).
- Add VTT normalization to extract spoken text from WebVTT files, removing timestamps, metadata, and punctuation to ensure the WER calculation reflects actual recognition quality.
- Add reference/baseline txt and vtt files for WER calculation, given output from a new integration test run (reference transcriptions were obtained from https://argo-stage.stanford.edu/view/druid:ys535zt3073, a result of this integration test; transcripts were validated by @jmartin-sul watching the video and reading the txt, as well as watching the video with the vtt selected as the caption source)
- Add the `AGENTS.md` file (and a `CLAUDE.md` symlink to it) that Gemini used in helping write this PR

Written by pairing with Gemini CLI and tweaking its work.  In particular, it came up with the WER mapping trick to use the distance gem we'd already settled on, as well as the VTT normalization approach.

ref https://github.com/sul-dlss/speech-to-text/issues/90 --
> The integration test passes but it doesn't do much with the resulting transcript other than ensure it's there and looks like it's about the right size.

Here's one attempt at fixing that part of the problem.

I had been worrying aloud lately about how we test model-based outputs on an ongoing basis in a meaningful way, and this seemed like a good opportunity to experiment with one relatively basic approach to the problem (and it was an opportunity to get some practice pairing with an AI agent on a somewhat unfamiliar problem space).

It doesn't seem like support for this in Ruby is great.  Another approach I considered (but didn't pursue) was using `uvx` to put a little Python script in this project, since the support for this sort of thing seems better in Python land (see e.g. [werpy](https://github.com/analyticsinmotion/werpy), a word error rate analysis package in Python).

I spent about a day on this at the end of last week, and I learned some things about Word Error Rate calculation, and WebVTT, and the related Ruby tooling for both, and got some Gemini practice.  So even if others see reasons this isn't a good idea and we close it, no big deal.

### DIFF
--- a/spec/features/preassembly_speech_to_text_media_spec.rb
+++ b/spec/features/preassembly_speech_to_text_media_spec.rb
@@ -150,8 +150,34 @@ RSpec.describe 'Create a media object via Pre-assembly and ask for it be speechT
 
     expect(files[10].text).to match(%r{video_log.txt text/plain 5\d* Bytes No role})
 
-    # TODO: Add expectations for the speech to text files when they are added to the object
-    #
+    # Download the generated transcription files and check their Word Error Rate (WER)
+    # against the human-validated reference files.
+    within(files[8]) { click_link_or_button 'audio_1_m4a.txt' }
+    # second link in the modal is the preservation download link, which should already
+    # be available (whereas the stacks link availability can lag accessioning a bit)
+    all('.modal-content a')[1].click
+    wait_for_download
+    txt_wer = calculate_wer('spec/fixtures/speech-to-text-reference-transcript.txt', download, format: :text)
+    puts " *** Word Error Rate (TXT): #{txt_wer} ***"
+    expect(txt_wer).to be < 0.05 # Allow for slight variations in ASR
+    delete_download(download)
+
+    # close the file modal
+    click_link_or_button 'Cancel'
+
+    within(files[9]) { click_link_or_button 'audio_1_m4a.vtt' }
+    # second link in the modal is the preservation download link, which should already
+    # be available (whereas the stacks link availability can lag accessioning a bit)
+    all('.modal-content a')[1].click
+    wait_for_download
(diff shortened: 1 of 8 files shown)


spec/features/preassembly_speech_to_text_media_spec.rb
```diff
@@ -150,8 +150,34 @@
...
+    # Download the generated transcription files and check their Word Error Rate (WER)
+    # against the human-validated reference files.
+    within(files[8]) { click_link_or_button 'audio_1_m4a.txt' }
+    # second link in the modal is the preservation download link, which should already
+    # be available (whereas the stacks link availability can lag accessioning a bit)
+    all('.modal-content a')[1].click
+    wait_for_download
+    txt_wer = calculate_wer('spec/fixtures/speech-to-text-reference-transcript.txt', download, format: :text)
+    puts " *** Word Error Rate (TXT): #{txt_wer} ***"
+    expect(txt_wer).to be < 0.05 # Allow for slight variations in ASR
```
edsu, line 162: Is it worth putting a comment here about what to do if the test fails? What would we expect a first responder to do? Personally I would prefer to do this checking in the speech-to-text repository...but I'm not going to stand in the way of this being merged!

edsu (approved):

peetucket (approved):
Pretty cool implementation.  Same comments as Ed, but I think we should merge and leave here and then could consider porting over to the STT project if/when we find it useful.

PR #1016, opened 2026-06-29
Update preassembly tests and include in test suite

aaron-collier:

Fixes to the preassembly tests to run in the new pattern.


spec/features/preassembly/preassembly_ocr_document_spec.rb
```diff
@@ -28,9 +27,13 @@
       # Check that ocrWF ran
       reload_page_until_timeout!(text: 'ocrWF')
 
-      # Wait for the second version accessioningWF to finish
-      reload_page_until_timeout_with_wf_step_retry!(expected_text: 'v2 Accessioned', workflow: nil) do |page|
-        if page.has_text?('v2 Accessioned')
+      # Wait for the new version accessioningWF to finish
+      elem = find_table_cell_following(header_text: 'Status')
+      md = /^v(\d+) *./.match(elem.text)
+      version = md[1].to_i
```
lwrubel, line 33: This is intended to loosen up which version it's checking on? Is that because we're re-using objects and so we can't know which version the test might be acting on?

aaron-collier: Exactly.

peetucket (approved):

PR #1023, opened 2026-07-08
verify is not needed?

peetucket:

As far as I can tell, the 'verify' version of reaccessioning spec is not needed, it just asserts the same thing that was in the re-accessioning spec?  This removes it, and clarifies ordering and dependencies in the README.

Yes, it was.


spec/features/preassembly/preassembly_reaccessioning_spec.rb
```diff
@@ -73,12 +73,16 @@
...
-    # Scroll to the bottom so the lazily-loaded section enters the viewport
-    # and the browser fetches its content.
-    page.scroll_to(:bottom)
+      # Scroll to the bottom so the lazily-loaded section enters the viewport
+      # and the browser fetches its content.
+      page.scroll_to(:bottom)
 
+      # events are loaded lazily, give the network a few moments
+      page.has_text?("v#{version} Accessioned", wait: 2)
+    end
```
peetucket, line 85: this is how it was done in the removed 'verify' spec

justinlittman (approved):

PR #1026, opened 2026-07-09
update reaccessioning spec

peetucket:

More updates to the re-accessioning spec:

- clearly indicate it does not work to re-run it if you try and stop since this will cause failures with expectations in weird ways (this spec is designed to run on a fresh object that was created by an earlier spec)
- tell the user how to re-run if needed (by deleting data files and rerunning just the specs needed to create a new object)
- update/add back-in expectations for the re-accessioned version of the object

Works full run and with re-runs (after following instructions).

Nope.

### DIFF
--- a/spec/features/preassembly/preassembly_reaccessioning_spec.rb
+++ b/spec/features/preassembly/preassembly_reaccessioning_spec.rb
@@ -53,12 +53,30 @@ RSpec.describe 'Create and re-accession image object via Pre-assembly', type: :p
   end
 
   scenario do
+    expect(page).to have_text(/v\d+ Accessioned/)
+
     # Get the original version from the page
     elem = find_table_cell_following(header_text: 'Status')
     md = /^v(\d+) Accessioned/.match(elem.text)
     version = md[1].to_i
 
-    expect(page).to have_text("v#{version} Accessioned")
+    if version > 1
+      raise <<~MESSAGE
+        This object is already above version 1, meaning it has already been re-accessioned by this spec before.
+        The spec assumes a freshly-accessioned object and is not safe to replay against an already-reaccessioned druid.
+
+        To get a fresh object if you want to run this spec again, remove the data and re-run the specs to create a new object:
+
+        ```
+        rm tmp/*_data.yml
+        bin/rspec spec/features/registration
+        bin/rspec spec/features/accessioning/preassembly_accessioning_spec.rb
+        bin/rspec spec/features/preassembly/preassembly_reaccessioning_spec.rb
+        ```
+
+         or just redo the full suite: `bin/rspec`.
+      MESSAGE
+    end
 
     files = all('tr.file')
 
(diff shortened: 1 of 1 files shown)


spec/features/preassembly/preassembly_reaccessioning_spec.rb
```diff
@@ -53,12 +53,30 @@
   end
 
   scenario do
+    expect(page).to have_text(/v\d+ Accessioned/)
+
```
peetucket, line 57: just move this assertion up so we can fail on the assertion if the object isn't accessioned, instead of when it tries to find the version number via a selector below, which is less clear when it happens

spec/features/preassembly/preassembly_reaccessioning_spec.rb
```diff
@@ -53,12 +53,30 @@
...
+        ```
+        rm tmp/*_data.yml
+        bin/rspec spec/features/registration
+        bin/rspec spec/features/accessioning/preassembly_accessioning_spec.rb
+        bin/rspec spec/features/preassembly/preassembly_reaccessioning_spec.rb
+        ```
+
+         or just redo the full suite: `bin/rspec`.
+      MESSAGE
```
peetucket, line 78: explain to the user you can't re-run the re-accession spec on an object that it's already been run on once , and tell them what to do

spec/features/preassembly/preassembly_reaccessioning_spec.rb
```diff
@@ -67,8 +85,8 @@
     expect(files[1].text).to match(%r{argo-logo.jp2 image/jp2 \d+\.*\d* KB})
     expect(files[2].text).to match(%r{image.jpg image/jpeg \d+.\d KB})
     expect(files[3].text).to match(%r{image.jp2 image/jp2 \d+\.*\d* KB})
-    expect(files[4].text).to match(%r{(?:sul-logo.png|vision_for_stanford.jpg) image/(?:png|jpeg) \d+.\d+ KB})
-    expect(files[5].text).to match(%r{(?:sul-logo|vision_for_stanford)\.jp2 image/jp2 \d+.\d+ KB})
+    expect(files[4].text).to match(%r{sul-logo.png image/png \d+.\d+ KB})
+    expect(files[5].text).to match(%r{sul-logo.jp2 image/jp2 \d+.\d+ KB})
```
peetucket, line 89: for v1 of the object, which is what we are asserting here, it should always be the sul-logo images

spec/features/preassembly/preassembly_reaccessioning_spec.rb
```diff
@@ -172,82 +190,38 @@
...
-
-    #   counter += 1
-    #   sleep sleep_duration
-    # end
-
-    # expect(image_response.status).to eq(200)
-    # expect(image_response.headers['content-type']).to include('image/jpeg')
-
-    # visit_argo_and_confirm_event_display!(druid:, version: latest_version)
-    # confirm_archive_zip_replication_events!(druid:, from_version: 1, to_version: latest_version)
```
peetucket, line 251: we are uncommenting/adding back in most of these assertions below (the ones in argo anyway to verify the object was versioned correctly) and ensuring the assertions are specific to the re-accessioned objects

spec/features/preassembly/preassembly_reaccessioning_spec.rb
```diff
@@ -172,82 +190,38 @@
...
+
+    # ensure changed files are all there, per pre-assembly
+    files = all('tr.file')
+    expect(files.size).to eq 6
+    expect(files[0].text).to match(%r{argo-logo.png image/png \d+.\d KB})
+    expect(files[1].text).to match(%r{argo-logo.jp2 image/jp2 \d+\.*\d* KB})
+    expect(files[2].text).to match(%r{image.jpg image/jpeg \d+.\d KB})
+    expect(files[3].text).to match(%r{image.jp2 image/jp2 \d+\.*\d* KB})
+    expect(files[4].text).to match(%r{vision_for_stanford.jpg image/jpeg \d+.\d+ KB})
+    expect(files[5].text).to match(%r{vision_for_stanford.jp2 image/jp2 \d+.\d+ KB})
```
peetucket, line 209: the re-accessioned object has different images, assert that

justinlittman (approved):

PR #1029, opened 2026-07-09
More integration test improvements

peetucket:

1. Indicate in README how to quickly re-run just the last failed specs (built into rspec, didn't know this).
2. Make selecting a collection when registering less flaky (account for collection drop-down populating dynamically when APO is selected).
3. Add the ability to quickly run a subset of accessioning specs, ignoring long running ones like OCR/captioning/GIS.  This is helpful if you want to just verify that the core accessioning pipeline is working relatively quickly.
4. Typo fixes

Yes.

Verified by running integration tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,22 @@ Each type of test can be run as a separate group with:
 bin/rspec --tag type:[registration|accessioning|sdr|verify]
 ```
 
+### Accessioning Quick Run
+
+To run a minimal end-to-end accessioning path (register an APO and collection, register all the standard test objects, accession a plain image object via Preassembly, then re-accession it) without the more time consuming H3/GIS/OCR/media accessioning specs, use the `sample_accession` tag:
+
+```
+bin/rspec --tag sample_accession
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -55,6 +55,22 @@ Each type of test can be run as a separate group with:
 bin/rspec --tag type:[registration|accessioning|sdr|verify]
 ```
 
+### Accessioning Quick Run
+
+To run a minimal end-to-end accessioning path (register an APO and collection, register all the standard test objects, accession a plain image object via Preassembly, then re-accession it) without the more time consuming H3/GIS/OCR/media accessioning specs, use the `sample_accession` tag:
+
+```
+bin/rspec --tag sample_accession
```
justinlittman, line 63: This is a super-helpful.

justinlittman (approved):

README.md
mjgiarlo: Agreed 👏🏻

